PR #11, opened 2015-01-05
Get status information via the zeo port and a nagios monitor

jimfulton:
https://groups.google.com/forum/#!topic/zodb/zMAzCXzilUI

### DIFF
--- a/src/ZEO/nagios.py
+++ b/src/ZEO/nagios.py
@@ -1,0 +1,141 @@
+from __future__ import print_function
+##############################################################################
+#
+# Copyright (c) 2011 Zope Foundation and Contributors.
+# All Rights Reserved.
+#
+# This software is subject to the provisions of the Zope Public License,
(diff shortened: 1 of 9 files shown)


tseaver:
OK, LGTM then.

src/ZEO/nagios.py
```diff
@@ -0,0 +1,145 @@
...
+    else:
+        m = re.match(r'(\S+):(\d+)$', addr)
+        if m:
+            addr = m.group(1), int(m.group(2))
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        else:
+            s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    try:
+        s.connect(addr)
+    except socket.error:
```
freddrake: Why not ``except socket.error as err``?  Does ZEO support Python older than 2.6?

jimfulton: Fixed.

PR #15, opened 2016-01-12
Python3: Use bytes, instead of strings on Windows too.

davidanes:
Seems that a leftover when upgading the code to support Python 3, the Unix version was updated but not the windows side.

UNIX sIde already uses bytes:

        def _physical_pull(self):
            os.write(self.trigger, b'x')

The error spitted if this patch is not applied is:

    Exception in thread Connect([(<AddressFamily.AF_INET: 2>, ['127.0.0.1', 4881])]):
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\threading.py", line 923, in _bootstrap_inner
        self.run()
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\client.py", line 390, in run
        success = self.try_connecting(attempt_timeout)
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\client.py", line 421, in try_connecting
        r = self._connect_wrappers(wrappers, deadline)
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\client.py", line 503, in _connect_wrappers
        wrap.connect_procedure()
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\client.py", line 589, in connect_procedure
        self.test_connection()
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\client.py", line 597, in test_connection
        self.conn = ManagedClientConnection(self.sock, self.addr, self.mgr)
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\connection.py", line 713, in __init__
        self.call_from_thread()
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\trigger.py", line 106, in pull_trigger
        self._physical_pull()
      File "E:\Dropbox\Projects\Python-x64-3.5.0\lib\site-packages\ZEO\zrpc\trigger.py", line 235, in _physical_pull
        self.trigger.send('x')
    TypeError: a bytes-like object is required, not 'str'

### DIFF
--- a/src/ZEO/zrpc/trigger.py
+++ b/src/ZEO/zrpc/trigger.py
@@ -21,7 +21,7 @@ import socket
 import errno
 
 from ZODB.utils import positive_id
-from ZEO._compat import thread, get_ident
+from ZEO._compat import thread, get_ident, ascii_bytes
 
 # Original comments follow; they're hard to follow in the context of
 # ZEO's use of triggers.  TODO:  rewrite from a ZEO perspective.
@@ -135,6 +135,9 @@ class _triggerbase(object):
     def __repr__(self):
         return '<select-trigger (%s) at %x>' % (self.kind, positive_id(self))
 
+# Added for Python 3.2 compatibility.
+TRIGGER_CHARACTER = ascii_bytes('x')
+
 if os.name == 'posix':
 
     class trigger(_triggerbase, asyncore.file_dispatcher):
@@ -159,7 +162,7 @@ if os.name == 'posix':
             asyncore.file_dispatcher.close(self)
 
         def _physical_pull(self):
-            os.write(self.trigger, b'x')
+            os.write(self.trigger, TRIGGER_CHARACTER)
 
 else:
     # Windows version; uses just sockets, because a pipe isn't select'able
@@ -232,4 +235,4 @@ else:
             self.trigger.close()
 
         def _physical_pull(self):
-            self.trigger.send(b'x')
+            self.trigger.send(TRIGGER_CHARACTER)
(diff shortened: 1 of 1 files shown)


src/ZEO/zrpc/trigger.py
```diff
@@ -232,4 +232,4 @@ def _close(self):
             self.trigger.close()
 
         def _physical_pull(self):
-            self.trigger.send('x')
+            self.trigger.send(b'x')
```
tseaver: We are still supporting Python 3.2, which doesn't have the bytes literal prefix.  You can use something like the following:

```python
from ZODB._compat import ascii_bytes

TRIGGER_CHARACTER = ascii_bytes('x')
```

and then use that constant instead.

mgedmin:
If [winbot](http://winbot.zope.org/builders/ZODB_dev%20py_270_win64) failed to notice the bug, perhaps we also need a new unit test?

mgedmin:
D'oh, winbot is not running any tests on Python 3.

I suggest we start using Appveyor.

agroszer:
winbot: sadly tox isn't helpful there

davidanes:
Fixed this with @tseaver comments both in posix and windows brachs of the code.

See https://github.com/davidanes/ZEO/commit/0d099c60082fd75c0f1a58fe842e40626bba06d5

davidanes:
Updated patch branch with commit in prev. comment, and update the pull request.

davidanes:
I don't know why this is failing now in the tests, will see later at home. We should review all code in zrpc for str/bytes and check that everything is right, IMHO.

PR #19, opened 2016-04-01
AttributeError 'server' when StorageServer creation fails

stefanholek:
When `StorageServer.__init__` fails with an exception, the error is masked by a subsequent AttributeError: 'ZEOServer' object has no attribute 'server'.

I propose to add a close_server function to ZEOServer, which can handle the case of non-existing server. This is also for the benefit of subclasses, which get better access to the server lifecycle.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ Changelog
 
 - Drop support for Python 2.6 and 3.2.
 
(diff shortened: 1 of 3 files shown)


mgedmin:
LGTM.

I've restarted the failed Travis build on Python 3.3, to see if that failure is reproducible.

PR #21, opened 2016-05-31
Asyncio-based ZEO client and server

jimfulton:
The main things in this change:

- Change from one asynchronous networking framework (asyncore) to
another (asyncio).

- Make connection logic much simpler.

- Simplify client threading model, eliminating many locks

- Reduce extra levels of indirection that made the code harder to reason about.

How to approach reviewing this?

I would recommend starting with ZEO.asyncio, noting that the cache is
now managed by the networking thread and doesn't require locks, then
look at the changes to ClientStorage, StorageServer, and ZEO.runzeo
and ZEO.tests.forker.  I would focus on one side at a time, client or server.

Ask questions in the PR or on IRC.

If you're feeling especially thorough, you could review the test changes.


jimfulton:
As of the latest build, the tests are all passing.  This is partly an accident. There are a few tests that fail intermittently under travis.  But no worries.  This PR is only the first installment of the asyncio work.  Notice that it merges to the ZEO5 branch, not to master.

jimfulton:
Now, I need a reviewer. :)

I'm working on the server side now. (I also realized that I need to be a little more careful with pickling and the server work will fix that.) If this PR hasn't been reviewed by the time the server PR is ready, I'll, um, "review" this PR myself.

jimfulton:
BTW, github is complaining about an earlier build.

jimfulton:
Heh.

Well... One could review just the new asyncio package and ClientStorage.  A lot of the changes are just test updates.

jamadden:
> 	Update to reflect lack of Py2 support

Is the plan to still try to use Trollius to bring back Py2 support, or is Py2 support (along with PyPy support) simply going to be dropped? That would make me very sad.

The mailing list had raised concern about the future of the Trollius project, and so some other alternatives (zc.ngi, IIRC, was one) were mentioned. Has there been more thought given to that? 

To my mind, it would be less painful to keep Py2 and Py3 working together rather than trying to re-support Py2 again later (in order to avoid accidentally depending on features that are difficult to backport), but that's just my barely-considered opinion.

jimfulton:
The plan is still to try to use trollius.  I even did a spike: https://github.com/zopefoundation/ZEO/tree/trollius
Lots of tests fail, but quite a few pass.  I'll get back to this.

While Trollius has been deprecated, its documentation says: 

"Since the Trollius is used for some projects in the wild, Trollius will not disappear. You can only expect minimum maintenance like minor bugfixes. Don’t expect new features nor synchronization with the latest asyncio."

I'm using very little of asyncio, so I'm hopeful we can limp along until 2020. :)  Plus there are plugins for it, which potentially reduce the support surface area further.

Python 2 is still important (although, at this point, also annoying :) ) to me.

jimfulton:
I should also note that this work:

- Drops the ZEO authentication protocol.  It will be replaced by something(s) else.  I expect probably some sort of cert authentication, as well as user/password (over ssl).

- Drops full cache verification.

- Drops support for old ZEO protocols.

jimfulton:
I ended up refactoring the asyncio client to share code with the server, so I want ahead and included the server in the PR.

jamadden:
I don't know enough about asyncio to feel comfortable as a reviewer of anything except coding style, sorry.

I did run zodbshootout on Python 3.4/OS X with ZEO 4.2.0 and this branch (594c8c9ba9b4be3945abfa6a83b2dcb2fa217846) for comparison (1000 objects in BTrees, concurrency=6; I didn't measure CPU or memory usage). I ran them several times and got similar results:

Test | ZEO 4.2 | ZEO asyncio | % speed (ZEO/asyncio)
------|-------------:|-------------------:|-------------------:
Add 1000 objects             | 12,201 | 9,328    | 76%
Update 1000 objects        | 10,754 | 8,651   | 80%
Read 1000 Warm objects |  4,958 | 5,477   | 110%
Read 1000 Cold objects   | 4,994 | 5,465    | 109%
Read 1000 Hot objects     | 75,194 | 25,137 | 33%
Read 1000 Steamin objects | 11,035,222 | 9,503,709   | 86%

jimfulton:
WRT asyncio, as the maintainer of gevent, you have a lot of knowledge on the domain, and learning a bit about asyncio wouldn't hurt you. :) <shrug> 

OTOH, I'm using asyncio in a traditional reactive style rather than the newfangled coroutine style, as that seems to be a better fit for the way ZEO works. (I've written lots of network apps in the coroutine style using zc.ngi, so I am a fan of it. I'm also a fan of gevent. :) )

WRT ZODB shootout, I'd done some initial analysis after doing the client implementation and found the asyncio implementation a little slower too. My test was done very informally on an EC2 t2.micro instance, I plan to get back to that and plan to do the test more formally and on a larger machine.

Your number above are suspect to me, especially the steamin hot tests which don't touch ZEO.

jimfulton:
I plan to work on debugging the travis failures today. All tests pass on my Mac.

jamadden:
>  learning a bit about asyncio wouldn't hurt you. :)

I completely agree, I should know more than I do. Maybe this will finally provide me the impetus to dig in.

> Your number above are suspect to me, especially the steamin hot tests which don't touch ZEO.

I should have asterisked that one. I agree it looks odd. I've seen a lot of unexplained variability there over my time with zodbshootout; even just running things now, the steamin number ping-pongs back and forth between the two implementations. I plan to run things again on a fresh reboot when I can get around to that. Still, though, the other numbers are pretty stable.

jimfulton:
Woo hoo, checks have passed.

An important internal change in this PR was to change the way test servers are run, and there was an issue with shutting them down on linux that didn't exist on Mac.

(There was also a long-existing bug that was made visible by a change in a race condition.)

PR #23, opened 2016-06-06
Load uses load before

jimfulton:
Changed loadBefore to operate more like load behaved, especially
with regard to the load lock.  This allows ZEO to work with the
upcoming ZODB 5, which used loadBefore rather than load.

Reimplemented load using loadBefore, this testing loadBefore
extensively via existing tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Changelog
   with regard to the load lock.  This allowes ZEO to work with the
   upcoming ZODB 5, which used loadbefore rather than load.
 
-  Reimplemented load using loadBefore, this testing loadBefore
+  Reimplemented load using loadBefore, thus testing loadBefore
   extensively via existing tests.
 
 - Fixed: the ZEO cache loadBefore method failed to utilize current data.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,15 @@ Changelog
 4.2.0 (unreleased)
 ------------------
 
+- Changed loadBefore to operate more like load behaved, especially
+  with regard to the load lock.  This allowes ZEO to work with the
+  upcoming ZODB 5, which used loadbefore rather than load.
+
+  Reimplemented load using loadBefore, this testing loadBefore
+  extensively via existing tests.
+
+- Fixed: the ZEO cache loadBefore method failed to utilize current data.
+
 - Drop support for Python 2.6 and 3.2.
 
 4.2.0b1 (2015-06-05)
--- a/src/ZEO/ClientStorage.py
+++ b/src/ZEO/ClientStorage.py
@@ -53,6 +53,9 @@ from ZODB import utils
 
 logger = logging.getLogger(__name__)
 
+# max signed 64-bit value ~ infinity :) Signed cuz LBTree and TimeStamp
+m64 = b'\x7f\xff\xff\xff\xff\xff\xff\xff'
+
 try:
     from ZODB.ConflictResolution import ResolvedSerial
 except ImportError:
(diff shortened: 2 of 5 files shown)


tseaver:
Travis shows failures on all supported python versions.

jimfulton:
Yeah, forgot to make a ZODB release before making this PR.

Rerunning travis.

jimfulton:
And rerunning didn't work because apparently, it didn't pick up the new zodb release.
I'll try again in a while.

jimfulton:
No, it installed 4.3.0. :(

jimfulton:
I'd screwed up the ZODB release, because I forgot to push the 4.3.0 release I'd made last week.
I made a ZODB 4.3.1 and reran travis build 127 and tests are passing.

tseaver:
Travis is still red. :(

jimfulton:
It wasn't in build 127. It went red because of reopening the pull request. ZEO tests are fragile, especially on travis.  I'll try  to run them again.

jimfulton:
Oh look! Travis is green now. :)

tseaver:
I've restarted build 126, too.

CHANGES.rst
```diff
@@ -4,6 +4,15 @@ Changelog
 4.2.0 (unreleased)
 ------------------
 
+- Changed loadBefore to operate more like load behaved, especially
+  with regard to the load lock.  This allowes ZEO to work with the
+  upcoming ZODB 5, which used loadbefore rather than load.
+
+  Reimplemented load using loadBefore, this testing loadBefore
+  extensively via existing tests.
```
tseaver: The "this testing" bit seems garbled:  typo?

src/ZEO/ClientStorage.py
```diff
@@ -53,6 +53,9 @@
 
 logger = logging.getLogger(__name__)
 
+# max signed 64-bit value ~ infinity :) Signed cuz LBTree and TimeStamp
+m64 = b'\x7f\xff\xff\xff\xff\xff\xff\xff'
```
tseaver: Should constants be all caps?  Also, this constant should be imported / used in the tests below which need the value.

CHANGES.rst
jimfulton: fixed

src/ZEO/ClientStorage.py
```diff
@@ -819,75 +822,37 @@ def load(self, oid, version=''):
...
-        try:
-            self._cache.store(oid, start, end, data)
-        finally:
-            self._lock.release()
+            if result:
+                with self._lock:    # for atomic processing of invalidations
+                    if self._load_status:
+                        data, tid, end = result
+                        self._cache.store(oid, tid, end, data)
+                    self._load_oid = None
```
tseaver: When / how does `self._load_oid` get reset in the case that `result` is False?  When/how does `self._load_status` get reset?

src/ZEO/cache.py
```diff
@@ -550,13 +553,22 @@ def load(self, oid):
     def loadBefore(self, oid, before_tid):
         noncurrent_for_oid = self.noncurrent.get(u64(oid))
         if noncurrent_for_oid is None:
-            self._trace(0x24, oid, "", before_tid)
-            return None
+            result = self.load(oid, before_tid)
+            if result:
+                return result[0], result[1], None
+            else:
+                self._trace(0x24, oid, "", before_tid)
```
tseaver: Wouldn't making the values passed as first arguments to `self._trace` named constants improve readability?

src/ZEO/ClientStorage.py
jimfulton: Well, z64 is lower case, and this is related.  I expect this will eventually migrate to ZODB.utils.

What tests need the value?

tseaver:
Apologies if the "peephole ignorance" of my comments makes them irrelevant:  reviewing something as complext as the `load`/`loadBefore` implementation without swapping everything back into RAM is hard.

jimfulton:
No problem. Thanks for reviewing. :)

src/ZEO/ClientStorage.py
tseaver: Hmm, I don't see it:  maybe my memory played tricks on me, and the constant was used in part of your changes to ZODB today?

src/ZEO/ClientStorage.py
jimfulton: That's a good question. You caught a bug.  I'll fix...

src/ZEO/cache.py
jimfulton: Probably, but that would be unrelated to this change.  The trace calls were just moved around a little.

src/ZEO/ClientStorage.py
jimfulton: Fixed.

Note that this was a pretty harmless bug, as _load_oid and _load_status would be cleared on the next load.

This load lock business is a mess. I'm glad it's going away.

tseaver:
Maybe a [race condition / resource starvation problem on Travis](https://travis-ci.org/zopefoundation/ZEO/jobs/135700097#L390-L400)?

src/ZEO/ClientStorage.py
jimfulton: No.  All I did there, for now, was to fix loadBefore, not implement load in terms of it.

jimfulton:
Travis is green for the build that github says is red.

Tres, are you good with this PR?

I have a tiny follow-on PR needed to make test pass with the ZODB 5 changes.

jimfulton:
Ok, look, making a comment made github poll travis. Cool. :)

tseaver:
Two hard topics in computer science:  cache invalidation, naming things, and off-by-one errors.

LGTM

src/ZEO/ClientStorage.py
jmuchemb: > I expect [m64] will eventually migrate to ZODB.utils.

We also have such value (named MAX_TID) in NEO because of SQLite (which does not accept numbers above 2^63-1). So that can be a good idea to specify it in ZODB, but we would not use it in NEO, because we want to limit the ZODB dependency to the client (tha's why we also redefine a few things from ZODB.utils).

PR #25, opened 2016-06-09
Zodb5

jimfulton:
These few changes make ZEO tests pass with both ZODB master and the no-more-load branch.

These changes are fixes and reflect the greater importance, now, of ``lastTransaction``.

### DIFF
--- a/src/ZEO/ClientStorage.py
+++ b/src/ZEO/ClientStorage.py
@@ -1119,7 +1119,8 @@ class ClientStorage(object):
             self._tpc_cond.release()
 
     def lastTransaction(self):
-        return self._cache.getLastTid()
+        with self._lock:
+            return self._cache.getLastTid()
 
     def tpc_abort(self, txn):
         """Storage API: abort a transaction."""
@@ -1473,10 +1474,11 @@ class ClientStorage(object):
             if oid == self._load_oid:
                 self._load_status = 0
             self._cache.invalidate(oid, tid)
-        self._cache.setLastTid(tid)
 
         if self._db is not None:
             self._db.invalidate(tid, oids)
+
+        self._cache.setLastTid(tid)
 
     # The following are for compatibility with protocol version 2.0.0
 
--- a/src/ZEO/ClientStorage.py
+++ b/src/ZEO/ClientStorage.py
@@ -1388,6 +1388,14 @@ class ClientStorage(object):
         for oid, tid in self._cache.contents():
             server.verify(oid, tid)
         server.endZeoVerify()
+
+        if server_tid > self._cache.getLastTid():
+            # We verified the cache, and got no new invalidations
+            # while doing so.  The records in the cache are valid,
+            # in that invalid current records were invalidated,
+            # but the last tid is wrong.  Let's fix it:
+            self._cache.setLastTid(server_tid)
+
         return "full verification"
 
     def invalidateVerify(self, oid):
--- a/src/ZEO/ClientStorage.py
+++ b/src/ZEO/ClientStorage.py
@@ -1390,12 +1390,13 @@ class ClientStorage(object):
             server.verify(oid, tid)
         server.endZeoVerify()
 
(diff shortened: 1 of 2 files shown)


jimfulton:
Travis is green. :)

jimfulton:
Is anyone going to review this?  If not, I'll go ahead and merge this today.

src/ZEO/ClientStorage.py
```diff
@@ -1388,6 +1389,14 @@ def verify_cache(self, server):
         for oid, tid in self._cache.contents():
             server.verify(oid, tid)
         server.endZeoVerify()
+
+        if server_tid > self._cache.getLastTid():
+            # We verified the cache, and got no new invalidations
+            # while doing so.  The records in the cache are valid,
+            # in that invalid current records were invalidated,
+            # but the last tid is wrong.  Let's fix it:
+            self._cache.setLastTid(server_tid)
```
jamadden: verify_cache is called at startup? 

Is there the possibility of invalidations being received while this method is running?  I'm wondering about race conditions possibly leading to setting the cache's lastTid *backwards*.

src/ZEO/ClientStorage.py
```diff
@@ -1465,11 +1474,12 @@ def _process_invalidations(self, tid, oids):
             if oid == self._load_oid:
                 self._load_status = 0
             self._cache.invalidate(oid, tid)
-        self._cache.setLastTid(tid)
 
         if self._db is not None:
             self._db.invalidate(tid, oids)
 
+        self._cache.setLastTid(tid)
+
```
jamadden: It's not clear to me why this moved down. Any particular reason?

src/ZEO/ClientStorage.py
jimfulton: ``verify_cache`` called on *connection* (see ``notifyConnected``) when the server tid is > cache tid and when "quick" verification isn't possible because we've been disconnected for too long.

Invalidations are queued while this ``verify_cache`` runs, however, the queueing stops just before the check above.  The logic is torturous in many ways, in part because of ZEO4's use of threads for managing connections, and because the logic is spread over too many methods. ZEO5 will simplify this quite a bit, including getting rid of full cache verification. :)

The race here is that ``server.endZeoVerify`` very indirectly calls ``endVerify``, which calls ``finish_verification``, which stops invalidation queueing before the ``server_tid > self._cache.getLastTid()`` check is done.  Bonus: the indirect call happens in a different client thread!

I'll fix this.  The fix will be a little tricky.  Need to think about it. Good catch!

src/ZEO/ClientStorage.py
jimfulton: Yes, to address the same race that Julien pointed out in the context of setting _txn_time in https://github.com/zopefoundation/ZODB/pull/56

If a connection called lastTransaction() after the cache tid was set but before getting the invalidations, then it would proceed with the transaction without having properly invalidating the cache.

This PR mitigates this race in 2 ways:

1. It gets the storage lock in ``lastTransactuon``.

2. It sets the cache tid (and thus the lastTransaction result) after sending invalidations.  The race is unlikely, but if we lose, we lose safely by using a transaction start time that's too early.

src/ZEO/ClientStorage.py
```diff
@@ -1388,6 +1389,15 @@ def verify_cache(self, server):
...
             server.verify(oid, tid)
         server.endZeoVerify()
+
+        with self._lock:
+            if server_tid > self._cache.getLastTid():
+                # We verified the cache, and got no new invalidations
+                # while doing so.  The records in the cache are valid,
+                # in that invalid current records were invalidated,
+                # but the last tid is wrong.  Let's fix it:
+                self._cache.setLastTid(server_tid)
```
jamadden: I see that `finish_verification` acquires and releases the lock, so this should close the possible race, and I don't see any possibility of deadlock.

jamadden:
Assuming travis is green again, LGTM now.

PR #29, opened 2016-06-19
Asyncio server

jimfulton:


tseaver:
Not worried about the Travis failures?

jimfulton:
I'm not worried about them here. I *am* worried about them in https://github.com/zopefoundation/ZEO/pull/21

which this PR got merged into.

I'll have to see if I can reproduce these on a slow linux ec2 instance.

PR #32, opened 2016-06-24
Add SSL support +

jimfulton:
Also:

- Change to a single-threaded server using ``asyncio.Server``. See the README.rst in the ``asyncio`` package. 

- Provide a new zconfig tag for ZEO clients.  In addition to modernizing settings, this also puts ZEO configuration in the ZEO package, where it belongs.

- Updates (more or less rewrites) the ZEO howto for accuracy and, I hope, greater usability.

- A few smaller cleanups, like removing documentation from legacy options, etc.


setup.py
```diff
@@ -92,7 +92,7 @@ def emit(self, record):
                     _unittests_only(suite, mod.test_suite())
     return suite
 
-tests_require = ['zope.testing', 'manuel', 'random2']
+tests_require = ['zope.testing', 'manuel', 'random2', 'mock']
```
michwill: should mock be in tox.ini as well?

jimfulton: I hope not. :) Dependencies should be fully specified in the setup.py.  If Tox required restating dependencies, that would be a strike against Tox.  I don't use Tox myself.  I'll use this as an excuse to try it and make sure it works.

jimfulton: Gawd -- it does require that I restate dependencies. Shame. "Fixing"

PR #43, opened 2016-07-08
Simplify server commit lock management

jimfulton:


jimfulton:
This needs a little more thought.

PR #45, opened 2016-07-12
Update zeo-client-cache-tracing to use accurate script names.

jamadden:
I haven't actually compared the output snippets (yet), but at least the script references are right now.

Here we're referring to full module paths. The alternative would be to add these scripts to the `entry-points` setup argument as zeostats and zeosimul.

### DIFF
--- a/doc/zeo-client-cache-tracing.txt
+++ b/doc/zeo-client-cache-tracing.txt
@@ -34,11 +34,12 @@ name or IP address) are logged.
 Analyzing a Cache Trace
 -----------------------
 
-The stats.py command-line tool is the first-line tool to analyze a cache
-trace.  Its default output consists of two parts:  a one-line summary of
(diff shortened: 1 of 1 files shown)


tseaver:
LGTM

jamadden:
The Python 2.7 Linux test has failed twice in a row in one test with "failed to start server"; all other tests passed. I can't see how that would be related.

jimfulton:
On Tue, Jul 12, 2016 at 11:54 AM, Jason Madden <notifications@github.com>
wrote:

> The Python 2.7 Linux test has failed twice in a row in one test with
> "failed to start server"; all other tests passed. I can't see how that
> would be related.
>
It's not. Just keep restarting that particular build (not the entire build)
till it goes green. :)

At some point, this will be annoying enough that I'll take the time to
figure out why the test servers sometimes fail to start.

jamadden:
It just went green :) Third times the charm.

PR #46, opened 2016-07-13
Convert cache_stats and cache_simul from getopt to argparse.

jamadden:
This results in much friendlier help and much improved argument parsing.


jamadden:
I can do this for the other scripts, if desired. I'm just personally using these two today for RelStorage, so that was my itch to scratch 😄

PR #48, opened 2016-07-14
Optimizations, featuring prefetch

jimfulton:
This PR has 3 optimizations:

- On load, check cache in calling thread rather than the networking thread. This improves cache read performance ~4x in my tests.

- Implemented a prefetch method, which allows you to get loads in progress asynchronously before loading them.  If you're good at predicting what you're going to load, then this can speeds significantly. (In my testing, this was about 60%, but would likely be more in a more realistic situation.)

- Collapse load requests. When there is an inflight loadBefore request, and the same request is made again, it uses the  inflight requests result rather than making a server call. This was needed to make prefetch effective.


jimfulton:
Note that to play with prefetch now (before it's added to ZODB, you can use:

    conn.db().storage.prefetch(oids, conn._strorage._start)

Yes, that's a lot of underware, but only for to experiment now. :)

src/ZEO/ClientStorage.py
```diff
@@ -504,8 +504,15 @@ def load(self, oid, version=''):
         return result[:2]
 
     def loadBefore(self, oid, tid):
+        result = self._cache.loadBefore(oid, tid)
+        if result:
+            return result
```
jamadden: So that makes it check the cache in the calling thread, just like ZEO 4 currently does. That seems safe and reasonable.

jamadden:
It's nice that prefetch can reuse the `loadBefore` protocol method. Since it takes an iterable of OIDs, did you consider if there was any benefit to a new protocol method that passes and returns the OIDs and data in bulk, for reduced network overhead? Or would that potentially be too large a message and interfere with the pipeline?

jimfulton:
I didn't think there would be much difference in network overhead.  The request sizes are small and the response payload dwarfs it's envelope, which is also small.  Note that because the request message id and the arguments are the same, I *did* consider avoiding sending both, but I resisted, so far. :)

The main benefit is in pipelining the requests, which you'd get either way.

I suspect the main benefit of adding fetch to the protocol would be that it would reduce the number of future's created in the client, except that those futures provide value, as mentioned below.

A potentially-major advantage of using ``loadBefore`` is that responses can come back earlier. When fetching many ids, the client could start populating the cache for the earlier oids before the server has processed the later ones.  Along the same lines, application code blocking on the earlier requests could be unblocked sooner.

Finally, the strategy of having clients reuse futures for inflight requests relies on having a future per ``loadBefore``.

BTW, I did a test to simulate a bulk load, which would be roughly equivalent to adding the protocol, and prefetching and letting the application start loading objects right away was ~30% faster. And of course, the effect would be amplified on a slower (non-localhost) network.

jamadden:
Thanks for writing that down. I had intuitively felt that the lower latency and shorter pipeline of more requests probably outweighed any request overhead benefits (and my typical default design pattern is to start with individual requests and only add a bulk option when it becomes apparent that it's necessary---individual requests are both easier to fan out across servers and cache, which is what the in-flight optimization is a form of). Your explanation solidifies it.

I'll be curious to see how this gets exposed at the ZODB layer.

So the concepts basically LGTM. I can't yet really comment on the details of the asyncio interactions because I haven't studied the base branch in detail (I find the `@self.protocol.XXX` decorators not immediately obvious without knowing more than I do).

jimfulton:
On Fri, Jul 15, 2016 at 6:30 PM, Jason Madden <notifications@github.com>
wrote:
...

> I'll be curious to see how this gets exposed at the ZODB layer.
>
Didn't you see this:
https://groups.google.com/forum/#!topic/zodb/RtpQiK5AYj0

It's odd. I thought this would generate some discussion.

> So the concepts basically LGTM. I can't yet really comment on the details
> of the asyncio interactions because I haven't studied the base branch in
> detail (I find the @self.protocol.XXX decorators not immediately obvious
> without knowing more than I do).
>
That makes me sad.

Hm.  I wonder if there's something I can do to make this clearer.

The ZEO implementation uses a callback style. I think that's a better fit
for the way that ZEO is used most of the time than, say, a coroutine
style.  Although, hm..... I need to think about this some more. The
built-in asyncio coroutines didn't fit my needs (even aside from Python 2),
but it might make sense to use ad-hoc generators to avoid callbacks.

Anyway, a callback API begs for anonymous functions, which Python lacks
(excluding lambdas, of course).  My typical work-around is to use
decorators.  Almost any time you see me use a decorator on an inner
function, the inner function is a callback.

If you're familiar with JavaScript promises:

    @self.protocol.promise(...)
    def func():
         ....

is roughly equivalent to:

    self.protocol.promise(...).then(func)

You've got me thinking about custom lightweight generators...

jamadden:
>> I'll be curious to see how this gets exposed at the ZODB layer.

> Didn't you see this:
https://groups.google.com/forum/#!topic/zodb/RtpQiK5AYj0

I did, but I was thinking more about implementation details. This isn't useful for some (synchronous) storages, so they can either not implement such a method or implement it as a no-op; so it can either be an optional part of the IStorage interface and checked for with `hasattr` or `try/except AttributeError` before calling, or they can be expected to implement it to do nothing, or it can be a new interface that's checked for, or something else...It's not especially important, I'm just curious 😃 .

> Anyway, a callback API begs for anonymous functions, which Python lacks
> (excluding lambdas, of course).  My typical work-around is to use
> decorators.  Almost any time you see me use a decorator on an inner
> function, the inner function is a callback.

I see. I would never have guessed. I found it confusing because it looks like a function definition just sitting there, doing nothing. It's rare to see decorators that have such *immediate, necessary side-effects*. I think i would prefer the imperative `then` style, myself, just because it *looks* like it's doing something.

In `@self.protocol.promise('loadBefore', oid, tid)` I can see what you mean, but that got changed to `@self.protocol.load_before(oid, tid)`, which I guess is a short-cut for the former? It lacks the `promise` wording though, so I wasn't sure. 

Again, I'm not trying to disparage the design. I just haven't read enough of the surrounding code to understand what's going on here from this snippet and the changes it contains.

jimfulton:
On Sat, Jul 16, 2016 at 9:17 AM, Jason Madden <notifications@github.com>
wrote:

> I'll be curious to see how this gets exposed at the ZODB layer.
>
> Didn't you see this:
> https://groups.google.com/forum/#!topic/zodb/RtpQiK5AYj0
>
> I did, but I was thinking more about implementation details. This isn't
> useful for some (synchronous) storages, so they can either not implement
> such a method or implement it as a no-op; so it can either be an optional
> part of the IStorage interface and checked for with hasattr or try/except
> AttributeError before calling, or they can be expected to implement it to
> do nothing, or it can be a new interface that's checked for, or something
> else...It's not especially important, I'm just curious 😃 .
>
Yes, it will be an optional duck-typed API. :)


> Anyway, a callback API begs for anonymous functions, which Python lacks
> (excluding lambdas, of course). My typical work-around is to use
> decorators. Almost any time you see me use a decorator on an inner
> function, the inner function is a callback.
>
> I see. I would never have guessed.
>
I understand. I've used this so long now, it's second nature to me. I even
threatened to implement transaction retry this way a while back.   I really
need to embrace the generator/coroutine model more, I guess.  Over the
years, after using JS and Scala a lot, lack of anonymous functions has
moved to the top of my Python peeve list.

> I found it confusing because it looks like a function definition just
> sitting there, doing nothing. It's rare to see decorators that have such *immediate,
> necessary side-effects*. I think i would prefer the imperative then
> style, myself, just because it *looks* like it's doing something.
>
Except the then style sucks without anonymous functions.

> In @self.protocol.promise('loadBefore', oid, tid) I can see what you mean,
>
As an aside, the main reason for Promise, aside from promise-like
invocation, is to make sure that promise callbacks are called immediately
rather than soon.  See the comment in client.Protocol.finish_connect if
you're curious.

> but that got changed to @self.protocol.load_before(oid, tid), which I
> guess is a short-cut for the former? It lacks the promise wording though,
> so I wasn't sure.
>
No, the issue there is that I had to use an asyncio.Future rather than a
Promise, because I needed to be able to have multiple done callbacks. In
the interest of simplicity, promises only allow one (well, a success
handler and/or an error handler).

> Again, I'm not trying to disparage the design.
>
No worries.

> I just haven't read enough of the surrounding code to understand what's
> going on here from this snippet and the changes it contains.
>
No one has, which is a big problem. I'm happy you're interested enough to
ask questions. Hopefully, my answers will lure you further in. :)

Jim

-- 
Jim Fulton
http://jimfulton.info

jimfulton:
I converted the code to use generators rather than promises.  I think the code is much cleaner and easier to follow now.

I'm annoyed that I didn't do this at first, but in my defense, asyncio's built-in generator support didn't fit my needs and I didn't have enough time with the API to realize that I could use short-lived generators for those cases where we weren't simply forwarding requests.

Of course this PR is bigger and does 2 things, but I felt it made sense in this case because of our discussion.

PR #51, opened 2016-07-18
Fixed: SSL clients of servers with signed certs didn't load default

jimfulton:
certs and were unable to connect.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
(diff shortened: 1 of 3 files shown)


jimfulton:
This PR also introduces a special value ``DYNAMIC`` for the ssl auth option.  I hadn't intended to so that in this PR but I think it's the hook I need right now. (And no one is going to review this PR anyway.)

jamadden:
> And no one is going to review this PR anyway.

 youch. But you may not be wrong.

PR #53, opened 2016-07-28
Added the ability to pass credentials when creating client storages.

jimfulton:
This is experimental in that passing credentials will cause
connections to an ordinary ZEO server to fail, but it facilitates
experimentation with custom ZEO servers. Doing this with custom ZEO
clients would have been awkward due to the many levels of
composition involved.

In the future, we expect to support server security plugins that
consume credentials for authentication (typically over SSL).

Note that credentials are opaque to ZEO. They can be any object with
a true value.  The client mearly passes them to the server, which
will someday pass them to a plugin.

### DIFF
--- a/src/ZEO/ClientStorage.py
+++ b/src/ZEO/ClientStorage.py
@@ -99,6 +99,7 @@ class ClientStorage(ZODB.ConflictResolution.ConflictResolvingStorage):
                  wait=True,
                  drop_cache_rather_verify=True,
                  username=None, password=None, realm=None,
+                 credentials=None,
                  # For tests:
                  _client_factory=ZEO.asyncio.client.ClientThread,
                  ):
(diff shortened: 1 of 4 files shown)


src/ZEO/ClientStorage.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, addr, storage='1', cache_size=20 * MB,
                  wait=True,
                  drop_cache_rather_verify=True,
                  username=None, password=None, realm=None,
```
michwill: If you have username and password in credentials, why do you also need them in keyword arguments?

jimfulton: These are left over from the old ZEO 4 authentication. I kept them for backward compatibility, but it's probably better to remove them. I'll do that in a separate PR.

PR #61, opened 2016-08-04
Client disconnect errors are now transient errors.

### LINKED ISSUE #56: Make ClientDisconnected retryable
Goal is that if we're disconnected, we don't return an error unless we remain disconnected for a long time.

To achieve the goal, we may need to did a little deeper. For example, we currently error immediately if we make an async call while disconnected, but maybe we should block. But maybe only certain methods, like tpc_begin...

jimfulton:
When applications retry jobs that raise transient errors, jobs (e.g. web
requests) with disconnect errors will be retried. Together with
blocking synchronous ZEO server calls for a limited time while
disconnected, this change should allow brief disconnections due to
server restart to avoid generating client-visible errors (e.g. 500
web responses).

Fixes #56

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: python
 sudo: false
+cache:
+  pip
+  directories:
+    - eggs
 matrix:
     include:
         - os: linux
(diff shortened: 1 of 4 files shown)


tseaver:
LGTM once Travis is happy.

Issue #58, opened 2016-08-03
Add an option to get ZEO to really sync.


jimfulton:
By making a server round trip.

jimfulton:
Fixed by https://github.com/zopefoundation/ZEO/pull/63

PR #62, opened 2016-08-04
Server sync

jimfulton:
Added a ClientStorage server-sync configuration option and
server_sync constructor argument to force a server round trip at
the beginning of transactions to wait for any outstanding
invalidations at the start of the transaction to be delivered.


jimfulton:
Fixes #58

PR #78, opened 2016-10-24
Avoid waiting on futures using timeouts -> major performance win on Python 2

jimfulton:
On Python 2, waiting on a future result with a timeout is very
expensive. (Because conditions with timeouts are very expensive.)

The thing is, we still want to use timeouts when we're disconnected.
We don't want requests to fail right away of the disconnection is
short.

We do a bit of a dance to only use a timeout if we're disconnected.

### DIFF
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -547,7 +547,7 @@ class Client(object):
     def get_peername(self):
         return self.protocol.get_peername()
 
-    def call_async_threadsafe(self, future, _, method, args):
+    def call_async_threadsafe(self, future, wait_ready, method, args):
         if self.ready:
             self.protocol.call_async(method, args)
             future.set_result(None)
@@ -557,7 +557,7 @@ class Client(object):
     def call_async_from_same_thread(self, method, *args):
         return self.protocol.call_async(method, args)
 
-    def call_async_iter_threadsafe(self, future, _, it):
+    def call_async_iter_threadsafe(self, future, wait_ready, it):
         if self.ready:
             self.protocol.call_async_iter(it)
             future.set_result(None)
@@ -623,7 +623,7 @@ class Client(object):
         except Exception:
             logger.exception("prefetch %r %r" % (oid, tid))
 
-    def prefetch(self, future, _, oids, tid):
+    def prefetch(self, future, wait_ready, oids, tid):
         if self.ready:
             for oid in oids:
                 if self.cache.loadBefore(oid, tid) is None:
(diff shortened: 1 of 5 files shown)


src/ZEO/asyncio/client.py
```diff
@@ -547,7 +547,7 @@ def verify(self, server_tid):
     def get_peername(self):
         return self.protocol.get_peername()
 
-    def call_async_threadsafe(self, future, method, args):
+    def call_async_threadsafe(self, future, _, method, args):
```
tseaver: Wouldn't we be better off naming the argument `wait_ready`?  Or do you have a linter which is complaining about unused arguments?

src/ZEO/asyncio/client.py
```diff
@@ -557,7 +557,7 @@ def call_async_threadsafe(self, future, method, args):
     def call_async_from_same_thread(self, method, *args):
         return self.protocol.call_async(method, args)
 
-    def call_async_iter_threadsafe(self, future, it):
+    def call_async_iter_threadsafe(self, future, _, it):
```
tseaver: Ditto for the `wait_ready` argument.

src/ZEO/asyncio/client.py
jimfulton: I prefer to use ``_`` to indicate an unused variable.  This is a common pattern and even built in to some languages.

jamadden: I also appreciate `_` for unused, but most linters I know will accept it as a prefix, i.e., `_some_unused`, which can meet both needs (indicating unused, while documenting).

jimfulton:
The new test failures are due to a race.  When doing load_before there's some code that looks like:

                future.set_result(data)
                if data:
                    data, start, end = data
                    self.cache.store(oid, start, end, data)

We notify the future before putting data in the cache. Some tests are effectively waiting on the future and are checking the cache size before the cache is updated.

What's especially weird:

- Only Python 2.7

- Not on MacOS

Some difference in thread scheduling I guess.

I could "fix" this by notifying the the future after updating the cache, but it seems slightly better to notify the future sooner, given that updating the cache potentially takes some meaningful time, because it's doing I/O.  

This makes the test code a little awkward. Someone might argue that as things are now, the cache data are inconsistent with the load result, but really the cache is an implementation detail, except when doing whitebox tests.

Thoughts?

jimfulton:
WRT the unused ``wait_ready`` argument, I prefer ``_``, but I'm willing to defer. I want people to review PRs. :) If you want me to change this, let me know.

jimfulton:
Tests passing, I changed _ -> wait ready.  LG?

PR #79, opened 2016-11-18
Fixed: ZEO needed changes to work with recent transaction changes.

jimfulton:


jimfulton:
I'll have a similar PR for the 4 branch.

src/ZEO/ClientStorage.py
```diff
@@ -799,40 +796,35 @@ def tpc_begin(self, txn, tid=None, status=' '):
             raise POSException.ReadOnlyError()
 
         try:
-            tbuf = txn.data(self)
+            tbuf = txn.__buf
         except AttributeError:
-            # Gaaaa. This is a recovery transaction. Work around this
```
jamadden: Recovery transactions are TransactionRecords, right? Perhaps it would be cleaner to give them the same data/set_data interface as a Transaction has? In the same way that we unified the extension/extended_info attribute between the two.

jimfulton: Except that the reason transactions need that is because there may be more than one data manager, but for transaction metadata and recovery transaction data, TCBOO client.

jimfulton: BTW, did you get a gitter notification from me? :)

jamadden:
> On Nov 18, 2016, at 09:32, Jim Fulton <notifications@github.com> wrote:
> 
> BTW, did you get a gitter notification from me? :)

Eventually, yes. Three lines rolled up into one email with a generic subject. Much less convenient than the notifications for github comments which come faster and individually, and have a specific subject and can be replied to from my email client.

jamadden:
> On Nov 18, 2016, at 09:31, Jim Fulton <notifications@github.com> wrote:
> 
> Except that the reason transactions need that is because there may be more than one data manager, but for transaction metadata and recovery transaction data, TCBOO client.

True enough. But then again, mucking around with Transaction's __dict__ is a bit smelly when there's an API provided for this purpose, so one way or the other one use-case is a loser. It seems nicer to me that objects that are meant to be duck-typed and used by the same methods actually quack alike; resorting to private attributes is an implicit acknowledgment that the APIs are incompatible or insufficient. (Plus, hypothetically, it prevents us from changing Transaction and/or TransactionRecord to use __slots__, too, were that to be a good idea.)

jimfulton:
OK, I don't feel strongly about this. I'm willing to data-fy TransactionMetaData objects. 

As far as the objects passed to tpc_begin in the context of recovery, there are potentially many implementations of these.  IDK that TransactionRecord is the only (or the base class of all) implementation of IStorageTransactionInformation.

jimfulton:
There's a ZODB PR to add the data API to both TransactionMetaData and TransactionRecord

jimfulton:
This PR will depend on it :)

jimfulton:
OK, updated and tests passing...

jimfulton:
Thanks

PR #80, opened 2016-11-18
Msgpack

jimfulton:
This PR adds optional msgpack support.

If you build the client and server using the msgpack extra and configure the server to use msgpack (via the config file or an environment variable), then msgpack will be used for the wire protocol. 

Why:

- byteserver in the future

- Python version independence of the ZEO server.  A Server running Python 3 can work with a
  client running Python 2.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+- Added support for serializing ZEO messages using `msgpack
+  <http://msgpack.org/index.html>`_ rather than pickle.  This helps
+  pave the way to supporting `byteserver
+  <https://github.com/jimfulton/byteserver>`_, but it also allows ZEO
+  servers to support Python 2 or 3 clients (but not both at the same
+  time) and may provide a small performance improvement.
+
 5.0.2 (2016-11-02)
 ------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,7 +20,6 @@ Changelog
 
 - Temporarily require non-quite-current versions of ZODB and
   transaction until we can sort out some recent breakage.
->>>>>>> origin/master
 
 5.0.2 (2016-11-02)
 ------------------
--- a/src/ZEO/asyncio/base.py
+++ b/src/ZEO/asyncio/base.py
@@ -127,7 +127,6 @@ class Protocol(asyncio.Protocol):
                     self.getting_size = True
                     self.message_received(collected)
             except Exception:
-                #import traceback; traceback.print_exc()
                 logger.exception("data_received %s %s %s",
                                  self.want, self.got, self.getting_size)
 
--- a/src/ZEO/asyncio/base.py
+++ b/src/ZEO/asyncio/base.py
@@ -10,8 +10,6 @@ import socket
 from struct import unpack
 import sys
 
-from .marshal import encoder
-
 logger = logging.getLogger(__name__)
 
 INET_FAMILIES = socket.AF_INET, socket.AF_INET6
@@ -129,13 +127,13 @@ class Protocol(asyncio.Protocol):
                     self.getting_size = True
                     self.message_received(collected)
             except Exception:
+                #import traceback; traceback.print_exc()
                 logger.exception("data_received %s %s %s",
                                  self.want, self.got, self.getting_size)
 
     def first_message_received(self, protocol_version):
         # Handler for first/handshake message, set up in __init__
         del self.message_received # use default handler from here on
-        self.encode = encoder()
         self.finish_connect(protocol_version)
 
     def call_async(self, method, args):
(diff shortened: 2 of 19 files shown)


jimfulton:
I wish someone would review this.

CHANGES.rst
```diff
@@ -13,6 +20,7 @@ Changelog
 
 - Temporarily require non-quite-current versions of ZODB and
   transaction until we can sort out some recent breakage.
+>>>>>>> origin/master
```
tiberiuichim: merge leftover?

jimfulton: Yup. Thanks.

src/ZEO/asyncio/base.py
```diff
@@ -129,13 +127,13 @@ def data_received(self, data):
                     self.getting_size = True
                     self.message_received(collected)
             except Exception:
+                #import traceback; traceback.print_exc()
```
jamadden: Unintended leftover.

src/ZEO/runzeo.py
```diff
@@ -342,6 +343,7 @@ def create_server(storages, options):
         storages,
         read_only = options.read_only,
         client_conflict_resolution=options.client_conflict_resolution,
+        msgpack=options.msgpack or os.environ.get('ZEO_MSGPACK'),
```
jamadden: So even if the user specifies `--msgpack=0` at the command line (assuming that's how the argument processing works), the environment variable will override? That seems backwards to me (explicit arguments should have precedence over implicit inherited environment; the other way around is hard to debug). Maybe have the default value of the argument be None, and only if the argument isn't specified at all look at the environment.

src/ZEO/asyncio/client.py
```diff
@@ -199,9 +205,12 @@ def finish_connect(self, protocol_version):
 
     exception_type_type = type(Exception)
     def message_received(self, data):
-        msgid, async, name, args = decode(data)
+        msgid, async, name, args = self.decode(data)
         if name == '.reply':
             future = self.futures.pop(msgid)
+            if isinstance(future, tuple):
+                future = self.futures.pop(future)
```
jamadden: It's not entirely clear to me why the handling of `self.futures` and tuples changed. Is that related to msgpack?

src/ZEO/asyncio/base.py
jimfulton: Yup. I'll remove.

(I wish Python has macros. I find myself adding and removing the same debugging code over and over. Sigh.)

src/ZEO/runzeo.py
jimfulton: IDK. ZConfig options handling is a mystery to me. I'll dig a little deeper.

Assuming that I can make it work the way you suggest, I will.

src/ZEO/asyncio/client.py
jimfulton: That's a good question. It's been so long, I don't remember for sure.

This is definitely wanted by the byte server so if not in this PR, then the next one.  Arguably, It makes the protocol cleaner as well.  Until fairly recently, msgid was always an integer. Then, when adding multiple simultaneous loads and prefetch, I added the hack of letting msgid be an (oid,tid) tuple. This hack worked because of the dynamic nature of Python. Rust, of course, is statically typed and it's a lot more convenient for msgid to always be an int. 

IIRC it *was* also helpful for msgpack, because by default, "arrays" are decoded as lists, which was messing up msgid lookup. I think I later change the default decoding for arrays to tuple, which would have made this a non-issue.

jamadden: Would it be possible to back this part of it out for now so that when it is needed it goes with a more logical PR?

I understand if you'd rather not back it out, and I won't complain. I just had to ask :)

src/ZEO/asyncio/base.py
jimfulton: done

src/ZEO/runzeo.py
jimfulton: Done.

jimfulton:
Thanks much for the review!

jamadden:
The Travis PR build failed for 3.4 due to an apparent hang:
```
Running .testZeo-new_addr.test tests:
  Running in a subprocess.
  Set up .testZeo-new_addr.test in 0.000 seconds.
  Ran 1 tests with 0 failures, 0 errors and 0 skipped in 0.518 seconds.
  Tear down .testZeo-new_addr.test in 0.000 seconds.
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

I've restarted that build...

jimfulton:
Thanks.

jimfulton:
Yes.   This is ZEO. Occasionally I wage war on individual intermittent failures and the number has decreased over time.  I'm probably due for another bout soon.

jamadden:
I have the same thing with gevent.

PR #86, opened 2017-04-13
Make all classes new-style. This is especially good for PyPy

jamadden:
See https://github.com/zopefoundation/ZODB/pull/160 for rationale.

Also restore PyPy to the build matrix.

I didn't try too hard to benchmark the actual improvements because there's a bug in the latest PyPy's (5.7.1) jit (I presume) that caused crashes (even with ZEO 4) under heavy loads (mostly on the client in ZEO5, mostly on the server in ZEO 4, interestingly enough). I did verify that under light load it at least performed no worse than before, both with and without zopefoundation/ZODB#160 being applied.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ matrix:
         - os: linux
           python: 2.7
         - os: linux
+          python: pypy-5.4.1
+        - os: linux
           python: 3.4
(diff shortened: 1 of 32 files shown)


jimfulton:
Thanks again.

PR #88, opened 2017-05-22
fixes #87: Attempted relative import in non-package

### LINKED ISSUE #87: runzeo error (Plone 5/Python 2.7): Attempted relative import in non-package
In order to not run into the problems of #84 I tried to use ``runzeo`` with supervisor. 

``runzeo`` fails to start with:

```
  File "/home/jensens/.buildout/shared-eggs/ZEO-5.1.0-py2.7.egg/ZEO/runzeo.py", line 401, in <module>
    main()
  File "/home/jensens/.buildout/shared-eggs/ZEO-5.1.0-py2.7.egg/ZEO/runzeo.py", line 392, in main
    s.main()
  File "/home/jensens/.buildout/shared-eggs/ZEO-5.1.0-py2.7.egg/ZEO/runzeo.py", line 155, in main
    self.create_server()
  File "/home/jensens/.buildout/shared-eggs/ZEO-5.1.0-py2.7.egg/ZEO/runzeo.py", line 251, in create_server
    self.server = create_server(self.storages, self.options)
  File "/home/jensens/.buildout/shared-eggs/ZEO-5.1.0-py2.7.egg/ZEO/runzeo.py", line 345, in create_server
    from .StorageServer import StorageServer
ValueError: Attempted relative import in non-package
```

I added a configuration to the Plone coredev buildout here:
https://github.com/plone/buildout.coredev/blob/5.1/experimental/zeoserver.cfg
and run it with ./bin/buildout -c experimental/zeoserver.cfg
The generated script ``./bin/zeoserver`` has two modes, one uses ``zeoctl`` and the other, if in enviroment ``SUPERVISOR_ENABLED=1`` is set, ``runzeo`` is used.

After changing the relative import to an absolute one this specific problem is gone and ``runzeo`` or ``SUPERVISOR_ENABLED=1 ./bin/zeoserver`` works (but #84 is still an issue).

jensens:
see #87

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,9 +4,13 @@ Changelog
 5.1.1 (unreleased)
 ------------------
 
+- Use absolute import in ``runzeo`` in order to avoid 
+  ``ValueError: Attempted relative import in non-package``, see
+  `issue #87 <https://github.com/zopefoundation/ZEO/issues/87>`_.
+
 - All classes are new-style classes on Python 2 (they were already
   new-style on Python 3). This improves performance on PyPy. See
-  `issue 86 <<https://github.com/zopefoundation/ZEO/pull/86>`_.
+  `issue 86 <https://github.com/zopefoundation/ZEO/pull/86>`_.
 
 
 5.1.0 (2017-04-03)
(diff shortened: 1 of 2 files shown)


tseaver:
I restarted the Python 2.7 Travis test build:  it looks like there may be a flaky test, unrelated to this change.

jamadden:
FWIW, I tend to agree with the comment on #87. In my experience (again, FWIW) this error (beyond the obvious) tends not to be the fault of the called package, but an error in the caller.

PR #91, opened 2017-07-11
runzeo clears socket files no matter what type of string address is

### LINKED ISSUE #90: ZConfig 3.2.0 leads to stale socket files on Python 2
[A change](https://github.com/zopefoundation/ZConfig/commit/010cb2b2d4d2cdc8cd327f9f46a987024cac5b67#diff-bd559d1f48a7d2df54fca55965db90b2R223) in ZConfig 3.2.0 leads to socket binding addresses being text on both Python2 and Python3.

This breaks runzeo.py's [`ZEOServer.clear_socket` method](https://github.com/zopefoundation/ZEO/blob/master/src/ZEO/runzeo.py#L196) (on Python 2) which only removes the socket file if it is the type of a string literal:
```python

def clear_socket(self):
  if isinstance(self.options.address, type("")):
     try:
        os.unlink(...)
```

Under Python 2, the string literal is a `bytes` object, not a text object.

Now, this wasn't a documented ZConfig change and it wasn't even exactly intended (it was the result of consistency improvements between the behaviour of Python 2 and Python 3, plus bug fixes). Still, because this project already has a dependency on `six`, it seems to make sense to use `six.string_types` there instead of `type("")`, which would solve the issue.

jamadden:
Fixes #90

Also remove an unused and duplicate import from that file and the redundant `test_suite` declaration from the test file. There was also a duplicate method definition in the test file that was removed.


jamadden:
Travis is reporting that the builds hang. I get the same hangs locally if I test under buildout with `bin/test -v1j99`, like what Travis uses. If I interrupt the process at the hang, I get this:
```
**********************************************************************
Could not communicate with subprocess!
Child command line: ['//bin/python3.6', '//ZEO/bin/test', '--resume-layer', '.SSLConnectionTests ZEO Connection Tests', '19', '--default', '--all', '--default', '--test-path', '--default', '//ZEO/src', '-v1j99']
Child stderr was:

**********************************************************************


**********************************************************************
Could not communicate with subprocess!
Child command line: ['//bin/python3.6', '//ZEO/bin/test', '--resume-layer', '.FileStorageSSLTests', '13', '--default', '--all', '--default', '--test-path', '--default', '//ZEO/src', '-v1j99']
Child stderr was:

**********************************************************************
```

If I run just `bin/test -vv` (so sequentially) I am starting to see all the FileStorageSSLTests time out:
```
Running .FileStorageSSLTests tests:
  Tear down .FileStorageReconnectionTests ZEO Connection Tests in 0.000 seconds.
  Set up .FileStorageSSLTests in 0.000 seconds.
  Running:
 check2StorageThreads (ZEO.tests.testZEO.FileStorageSSLTests)

Error in test check2StorageThreads (ZEO.tests.testZEO.FileStorageSSLTests)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 597, in run
    self.setUp()
  File "//ZEO/src/ZEO/tests/testZEO.py", line 172, in setUp
    **self._client_options()),
  File "//ZEO/src/ZEO/ClientStorage.py", line 279, in __init__
    self._wait()
  File "//ZEO/src/ZEO/asyncio/client.py", line 825, in wait
    self.wait_for_result(self.client.connected, timeout)
  File "//ZEO/src/ZEO/asyncio/client.py", line 759, in wait_for_result
    raise ClientDisconnected("timed out waiting for connection")
ZEO.Exceptions.ClientDisconnected: timed out waiting for connection

 check2ZODBThreads (ZEO.tests.testZEO.FileStorageSSLTests)

Error in test check2ZODBThreads (ZEO.tests.testZEO.FileStorageSSLTests)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 597, in run
    self.setUp()
  File "//ZEO/src/ZEO/tests/testZEO.py", line 172, in setUp
    **self._client_options()),
  File "//ZEO/src/ZEO/ClientStorage.py", line 279, in __init__
    self._wait()
  File "//ZEO/src/ZEO/asyncio/client.py", line 825, in wait
    self.wait_for_result(self.client.connected, timeout)
  File "//ZEO/src/ZEO/asyncio/client.py", line 759, in wait_for_result
    raise ClientDisconnected("timed out waiting for connection")
ZEO.Exceptions.ClientDisconnected: timed out waiting for connection


 check4ExtStorageThread (ZEO.tests.testZEO.FileStorageSSLTests)
```

I don't believe that's related to this change.

jamadden:
I debugged it to the point that I could see the client connecting to the server over and over but the connection failed. I had to drop back to Python 2 to determine why. Python 3 was just reporting a useless "ConnectionResetError", but Python 2 helpfully gives me this (on the server side):
```
  File "//ZEO/eggs/trollius-2.1-py2.7.egg/trollius/selector_events.py", line 230, in _accept_connection2
    yield From(waiter)
  File "//ZEO/eggs/trollius-2.1-py2.7.egg/trollius/selector_events.py", line 817, in _on_handshake
    wrap_ssl_error(self._sock.do_handshake)
  File "//ZEO/eggs/trollius-2.1-py2.7.egg/trollius/py3_ssl.py", line 149, in wrap_ssl_error
    return func(*args, **kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py", line 840, in do_handshake
    self._sslobj.do_handshake()
SSLError: [SSL: SSLV3_ALERT_CERTIFICATE_EXPIRED] sslv3 alert certificate expired (_ssl.c:661)
```

``client.pem`` expired June 23 of this year, as did ``server.pem`` and ``serverpw.pem``. So those files need to be regenerated for this to work again.

jimfulton:
hahaha
whimper

jamadden:
I've pushed a set of updated certs. Fingers crossed for Travis.

jamadden:
Travis is green.

src/ZEO/runzeo.py
```diff
@@ -193,11 +191,12 @@ def can_connect(self, family, address):
             return 1
 
     def clear_socket(self):
-        if isinstance(self.options.address, type("")):
+        if isinstance(self.options.address, six.string_types):
             try:
                 os.unlink(self.options.address)
             except os.error:
                 pass
+            return True
```
tseaver: Why are we returning `True` here?  And I don't see test coverage below for the branch where the address is *not* a string.

src/ZEO/tests/testZEOServer.py
```diff
@@ -138,8 +135,18 @@ def testFailCreateServer(self):
...
+
+        server = ZEOServer(Options())
+        self.assertTrue(server.clear_socket())
+
+    def test_clear_with_unicode_str(self):
+        class Options(object):
+            address = u"a str that does not exist"
+
+        server = ZEOServer(Options())
+        self.assertTrue(server.clear_socket())
```
tseaver: As noted above, we should be testing for the non-Unix-domain-socket case.

src/ZEO/runzeo.py
jamadden: I needed something to be able to distinguish that we took this branch in the test case.

And you're right, I will add a test for the opposite.

jamadden: Ah, I see that we have a test dependency on `mock`. I should be able to use that instead of a return.

jamadden:
Travis is green again after the requested changes. (I had to run some of the builds more than once due to apparent race conditions.)

jamadden:
Thanks! A PyPI release would be helpful when convenient (we're currently monkey-patching to get around this problem).

PR #94, opened 2017-07-11
Use python_requires and extras_require for version-specific deps

### LINKED ISSUE #92: Use extras_require for Python 2 dependencies
This will let us produce universal wheels.

### LINKED ISSUE #93: Use python_requires keyword for declarative version checking
Instead of imperatively doing so in setup.py. E.g., [pip](https://github.com/pypa/pip/blob/master/setup.py#L79). This will let PyPI know what python versions we run on, and will let pip and other tools that support that metadata (maybe that's baked into setuptools itself? I'm not sure) choose the appropriate version to install. This is also necessary to be able to ship wheels (because setup.py doesn't run).

jamadden:
(I don't actually expect this to work until #91 is merged because of the SSL test problems.)

Fixes #93.
Fixes #92 

Example:

```
$ pip install .
Processing /.../ZEO
ZEO requires Python '>=2.7.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*' but the running Python is 3.3.6`
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ Changelog
 
 - All classes are new-style classes on Python 2 (they were already
   new-style on Python 3). This improves performance on PyPy. See
-  `issue 86 <<https://github.com/zopefoundation/ZEO/pull/86>`_.
+  `issue 86 <https://github.com/zopefoundation/ZEO/pull/86>`_.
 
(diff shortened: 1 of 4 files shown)


jamadden:
Rebased and merged. Thanks!

PR #96, opened 2017-09-05
fix ZopeUndo.Prefix decode on python 2

mamico:
On python 2, calling undoLog via ZEO with ZopeUndo.Prefix filter, ZEO server raise an exception:

```
File ".../src/ZEO/asyncio/marshal.py", line 114, in pickle_server_decode
    return unpickler.load() # msgid, flags, name, args
File ".../src/ZEO/asyncio/marshal.py", line 164, in server_find_global
    raise ImportError("import error %s: %s" % (module, msg))
ImportError: import error copy_reg:
```

The problem was introduced when ZopeUndo.Prefix was changed and subclassed from the object.

This PR tries to fix the problem.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,6 @@ tests_require = [
     'random2',
     'mock',
     'msgpack-python',
-    'ZopeUndo',
 ]
 
 classifiers = """
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ tests_require = [
     'random2',
     'mock',
     'msgpack-python',
+    'ZopeUndo',
 ]
 
 classifiers = """
--- a/src/ZEO/tests/test_marshal.py
+++ b/src/ZEO/tests/test_marshal.py
@@ -7,7 +7,10 @@ from ZopeUndo.Prefix import Prefix
 class MarshalTests(unittest.TestCase):
 
     def testServerDecodeZopeUndoFilter(self):
-        # this is an example of Zope's 
+        # this is an example (1) of Zope2's arguments for
+        # undoInfo call. Arguments are encoded by ZEO client
+        # and decoded by server. The operation must be idempotent.
+        # (1) https://github.com/zopefoundation/Zope/blob/2.13/src/App/Undo.py#L111
         args = (0, 20, {'user_name': Prefix('test')})
         # test against repr because Prefix __eq__ operator
         # doesn't compare Prefix with Prefix but only
--- a/src/ZEO/tests/test_marshal.py
+++ b/src/ZEO/tests/test_marshal.py
@@ -1,11 +1,18 @@
 import unittest
 from ZEO.asyncio.marshal import encode
 from ZEO.asyncio.marshal import pickle_server_decode
-from ZopeUndo.Prefix import Prefix
+
+try:
+    from ZopeUndo.Prefix import Prefix
+except ImportError:
+    _HAVE_ZOPE_UNDO = False
+else:
+    _HAVE_ZOPE_UNDO = True
 
 
 class MarshalTests(unittest.TestCase):
 
+    @unittest.skipUnless(_HAVE_ZOPE_UNDO)
     def testServerDecodeZopeUndoFilter(self):
         # this is an example (1) of Zope2's arguments for
         # undoInfo call. Arguments are encoded by ZEO client
@@ -25,4 +32,3 @@ def test_suite():
     suite = unittest.TestSuite()
     suite.addTest(unittest.makeSuite(MarshalTests))
     return suite
-
--- a/src/ZEO/tests/test_marshal.py
+++ b/src/ZEO/tests/test_marshal.py
@@ -1,0 +1,25 @@
+import unittest
+from ZEO.asyncio.marshal import encode
+from ZEO.asyncio.marshal import pickle_server_decode
+from ZopeUndo.Prefix import Prefix
+
+
+class MarshalTests(unittest.TestCase):
+
+    def testServerDecodeZopeUndoFilter(self):
+        # this is an example of Zope's 
+        args = (0, 20, {'user_name': Prefix('test')})
+        # test against repr because Prefix __eq__ operator
+        # doesn't compare Prefix with Prefix but only
+        # Prefix with strings. see Prefix.__doc__
+        self.assertEqual(
+            repr(pickle_server_decode(encode(*args))),
+            repr(args)
+        )
+
+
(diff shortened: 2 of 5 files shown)


jensens:
any update here? this fixes a bug and is a blocker for the Plone 5.1 release.

src/ZEO/tests/test_marshal.py
```diff
@@ -0,0 +1,25 @@
...
+    def testServerDecodeZopeUndoFilter(self):
+        # this is an example of Zope's 
+        args = (0, 20, {'user_name': Prefix('test')})
+        # test against repr because Prefix __eq__ operator
+        # doesn't compare Prefix with Prefix but only
+        # Prefix with strings. see Prefix.__doc__
+        self.assertEqual(
+            repr(pickle_server_decode(encode(*args))),
+            repr(args)
+        )
```
tseaver: How does this test exercise the `copy_reg` / `__bulitins__` bits being added as the "fix"?

mamico: I added some information in the test comment.
Note: Probably the problem is only shown for the Zope version 2 branches, I see that on Zope4 ZopeUndo is completely eliminated (currently used by last Plone),

jensens: afaik in Zope4 Undo is still there, but was moved into the control panel and isn't available on most ZMI tabs.

icemac:
What ist the status of this PR? What needs to be done to get it merged?

jimfulton:
FTR, it pains me to introduce a dependency on ZopeUndo.  It pains me a lot.  I think we need a different way to approach this that involves a transaction-query language using standard data types.  I was prepared to do this a while ago when msgpack support was added, but there didn't seem to be interest.  I don't have time now.

I've been trying to ignore this PR and won't stand in it's way, but I would really prefer rethinking it to avoid the dependence on anything who's name has "Zope" in it.

mamico:
@jimfulton there is no new dependency to ZopeUndo here, the dependency in setup.py is only for the test requires. But +1 if you have a better solution for the regression bug.

src/ZEO/tests/test_marshal.py
```diff
@@ -0,0 +1,28 @@
+import unittest
+from ZEO.asyncio.marshal import encode
+from ZEO.asyncio.marshal import pickle_server_decode
+from ZopeUndo.Prefix import Prefix
```
tseaver: Can we make the import "soft", and drop the hard testing dependency?  E.g.:

```python
try:
    from ZopeUndo.Prefix import Prefix
except ImportError:
    _HAVE_ZOPE_UNDO = False
else:
    _HAVE_ZOPE_UNDO = True
```

and then:

```python
    @unittest.skipUnless(_HAVE_ZOPE_UNDO)
    def testServerDecodeZopeUndoFilter(self):
        ...
```

jensens: FTR: In Plone, we use the other approach checking for the package with ``pkg_resources``as documented in the [Style Guide under # 4](https://docs.plone.org/develop/styleguide/python.html#about-imports). I am not sure if this is needed in this case.

src/ZEO/tests/test_marshal.py
pbauer: @tseaver done

setup.py
```diff
@@ -33,6 +33,7 @@
     'random2',
     'mock',
     'msgpack-python',
+    'ZopeUndo',
```
jensens: That import is no longer used after ce623c3b6ede65f7db69aa32dbae55de92582be3

tseaver:
After we drop `ZopeUndo` from the test requirements, we may need to adjust `.travis.yml`, `tox.ini`, or `buildout.cfg` to ensure the package is still present.

setup.py
mamico: removed

jensens:
Travis CI uses buildout, so ``tox.ini`` and ``buildout.cfg`` only are affected.

jensens:
I changed this. I also found a problem with the ``unittest.skipUnless`` which I fixed.

Local tests are green ``Total: 1004 tests, 0 failures, 0 errors and 0 skipped in 6 minutes 44.312 seconds.`` 
I also activated Travis for PRs.

jensens:
On Travis, there are two failing tests with Python 3.4 and 3.5, but it looks unrelated to me. I have no idea whats going on there. 

Any ideas?

tseaver:
@jensens the Travis [branch](https://travis-ci.org/zopefoundation/ZEO/branches), [build history](https://travis-ci.org/zopefoundation/ZEO/builds), and [pr](https://travis-ci.org/zopefoundation/ZEO/pull_requests) pages show a lot of red over time.

My take is that the ZEO test suite takes too long to run, and therefore developers don't run it themselves enough.  E.g., last time I worked on ZEO, they ran long enough that I ran just the tests I knew I was working on: waiting 5 - 8 minutes for each run after an incremental change made doing the actual work impossible.

jensens:
@tseaver yes, it takes way to long. And getting the tox way of testing working is a lot of work with all the Python versions in there. But anyway, shouldn't it be this branch be green, as the master branch is?

jimfulton:
Some notes on testing issues.

1. A big problem is flaky tests.  I usually end up rerunning these to make sure they sometimes pass. 

   I recently learned about: https://pypi.python.org/pypi/flaky. I think we should use this for our tests that fail intermittently.  I think this will help a lot with Travis redness and for development, as it will be easier to get through the test suite.

   Of course, motivated contributors should occasionally work on flaky tests, which often result from testing issues, but occasionally reveal subtle real bugs.

2. I usually run with -j99. On my fairly old (late 2013) Mac, tests typically take < 3 minutes.

jimfulton:
FTR, I don't bother with tox, given that we have Travis.  IMO, it is sufficiently virtuous to get all tests passing in one environment and let Travis reveal the occasional  exception.

I think using the flaky tool will go a long way in making this process less frustrating.

jensens:
Just because I am curious what happens, I restarted the Python 3.4 job on master at https://travis-ci.org/zopefoundation/ZEO/jobs/252499407

jensens:
Ok, the same problem on the master. At least it is not a problem with this Pull Request.

pbauer:
I would love to merge this now? The test failure in the master-branch are unrelated to this PR. Any objections?

tseaver:
I'm fine with merging with unrelated Travis failures.

PR #103, opened 2018-01-19
Add coverage for ZEO module convenience functions

### LINKED ISSUE #102: mistake in writing
`ZEO/src/ZEO/__init__.py`

line 43 `ra
`should be `raise
`

tseaver:
Closes #102 after provoking it with a test.

### DIFF
--- a/src/ZEO/tests/testZEO.py
+++ b/src/ZEO/tests/testZEO.py
@@ -75,6 +75,30 @@ class CreativeGetState(persistent.Persistent):
         return super(CreativeGetState, self).__getstate__()
 
 
+class DummyClientThread(object):
+    def __init__(self, *args, **kw):
+        self._args = args
+        self._kw = kw
+    def call(self):
+        pass
+    def async(self):
+        pass
+    def async_iter(self):
+        pass
+    def wait(self):
+        pass
+
+
+class Test_convenience_functions(unittest.TestCase):
+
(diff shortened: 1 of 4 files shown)


src/ZEO/tests/testZEO.py
```diff
@@ -1636,7 +1724,9 @@ def pack(self, t=None, referencesf=None):
         ZEO.ClientStorage._check_blob_cache_size(self.blob_dir, 0)
 
 def test_suite():
-    suite = unittest.TestSuite()
+    suite = unittest.TestSuite((
```
jamadden: Is `suite = unittest.defaltTestLoader.loadTestsFromName(__name__)` an option here to avoid repetition?

tseaver: Heh, have look at that `test_suite` function -- I don't have enough in my gumption reservoir to do anything other than the most minimal change to get my tests to run.

tseaver:
> There are test failures in two environments, but they seem unrelated and have been happening for awhile

They are tests for @jimfulton's mta_acceptor / msgpack stuff, which I don't grok at all.

src/ZEO/tests/testZEO.py
jamadden: Hehe, fair enough. I didn’t want to deal with it either, I was just trying to pass the buck 😄

PR #110, opened 2018-03-27
Rename all 'async' to 'async_' for compatibility with Python 3.7.

jamadden:
Fixes #104.

I don't think any of those attributes were part of ZEO's public API.

PR #111, opened 2018-03-27
Disallow using strings as addresses on Windows. Fixes #106.

### LINKED ISSUE #106: ZEO ClientStorage wait_timeout is ignored on Windows
I was writing unit tests for zodbbrowser when I noticed that under Windows the following code never times out:

```python
from ZEO.ClientStorage import ClientStorage
storage = ClientStorage('/no/such/zeo/socket', wait_timeout=0.001, strorage='1', read_only=True)
```

Instead it hangs forever (I killed the hung Jenkins job after 19 hours).

The timeout works fine on Linux.

jamadden:
No specific test case, though.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@ Changelog
 - Fix ``ZEO.server`` relying on test dependencies. See `issue 105
   <https://github.com/zopefoundation/ZEO/issues/105>`_.
 
+- Disallow passing strings as addresses to ClientStorage under Windows
+  because string addresses are used for unix-domain sockets, which
+  aren't supported on Windows. See `issue 107
+  <https://github.com/zopefoundation/ZEO/issues/107>`_.
 
 5.1.2 (2018-03-27)
 ------------------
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 5.1.2 (unreleased)
 ------------------
 
+- Disallow passing strings as addresses to ``ClientStorage`` under
```
jimfulton: The main point of this isn't "strings". The point is that strings are used for unix-domain sockets, which aren't supported on windows.

Please reword, along the lines of:

- Disallow passing strings as addresses to ``ClientStorage`` under Windows because string addresses are used for unix-domain sockets, which aren't supported on Windows.

jamadden: OK, sure, that makes sense :) I was a little fast and loose with the change notes because I knew that there would be merge conflicts and they would probably need reworked anyway between #110 #112 and #111 . It's just down to which one gets merged first.

PR #114, opened 2018-03-28
Use pickle protocol 3 on the wire.

jamadden:
I also snuck the fix for ZEO_MTACCEPTOR tests in (use IPv4 instead of v6), so we have green builds (🙌 ).

On Python 2, this will necessitate that the client and server are both running at least this version of ZEO (because before this, ZEO wasn't using zodbpickle to read the wire so they won't be able to decode the protocol). Or use msgpack.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-5.1.3 (unreleased)
+5.2.0 (unreleased)
 ------------------
 
 - Fix ``ZEO.server`` relying on test dependencies. See `issue 105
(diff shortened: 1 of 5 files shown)


jamadden:
> I'm not sure the ZEO4 support tests are sane anymore with this change. I'm also not sure if that matters.

I also wondered about that and if it mattered. But the ZEO4 tests *did* require the changes to be able to read `zodbpickle.binary` objects, so they were already in some odd state.

jamadden:
Thank you! Could we get a PyPI release please?

jimfulton:
Done. 5.2.0

jamadden:
Thanks!

PR #115, opened 2018-05-31
'msgpack-python' egg was renamed 'msgpack'

jmuchemb:
https://github.com/msgpack/msgpack-python/#very-important-notes-for-existing-users

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ Changelog
 
 - Add support for Python 3.7.
 
+- Switch from `msgpack-python` to `msgpack`. Currently a version < 0.6
+  is required.
+
 
 5.2.0 (2018-03-28)
 ------------------
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ tests_require = [
     'manuel',
     'random2',
     'mock',
-    'msgpack-python',
+    'msgpack',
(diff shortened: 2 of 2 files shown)


mgedmin:
Should I retry the failed Python 3.6 tests to see if they're non-deterministic failures?  Or is anyone planning to analyze the failure and would rather I didn't destroy the current logs?

(I don't think Travis can let you see the logs of jobs that you've retried.  If I'm mistaken, I'd love to hear a correction!)

icemac:
@jmuchemb Could you please rebase your PR onto master, as it has now successfully running tests again.

icemac:
Using a `msgpack` version < 0.6 fixes the tests.
Can we merge this PR having a follow-up issue to update to the current version or should the issues be fixed in this PR? Any preference?

jamadden:
If the failures are simple and the fixes obvious, I think it'd be fine to fix them here while also updating the pin. 

However, from glancing at the log, neither the failures nor the fixes look simple (to me). In that case, it seems reasonable to merge this simple PR, so that master is working/releasable, and defer the version update to a later PR (so long as there's an issuer so we don't forget).

icemac:
I created #122 for the update to a current `msgpack` version.
Should this change have a change log entry or is it now ready to merge?

PR #120, opened 2018-11-25
Fix tests under ZODB 5.5

jamadden:
Based on discussion beginning https://github.com/zopefoundation/ZODB/pull/222#issuecomment-441398381

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,9 @@ install_requires = [
 ]
 
 tests_require = [
+    # We rely on implementation details of
+    # test mocks. See https://github.com/zopefoundation/ZODB/pull/222
+    'ZODB >= 5.5.1',
     'zope.testing',
     'manuel',
     'random2',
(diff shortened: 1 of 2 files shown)


jamadden:
I agree it is complicated. Thanks for the approval anyway!  A lot of the ugliness, I think, is due to unwrapping the decorated function on python 2. That will eventually go away, or go away even sooner if a future ZODB release sets the appropriate attribute.

jamadden:
> I have a feeling that reusing tests this way isn't as great an idea as it seemed at the time. It seems like it makes things easier, but it ends up making things kinda complicated, slow, and brittle.

I admit to having similar thoughts. But there is also a truly tremendous amount of value in having some sort of shared test specification for conformance to a standard of functionality across implementations. And when those implementations tend to use inheritance, sharing test classes is a natural mechanism. Perhaps things could be refactored differently, but the need for a standard test suite won’t go away.

Issue #140, opened 2019-04-25
Cannot run tests locally


jugmac00:
I tried to run the tests for ZEO locally, both via tox or directly via bin/test after manual buildout, and I get a lot of failures and errors.

On a first glance, there are two categories
- failures on asserts
- time out errors

Examples:
```
(ZEO) jugmac00@jugmac00-XPS-13-9370:~/Projects/ZEO$ bin/test -vv
/home/jugmac00/Projects/ZEO/src/ZEO/tests/testZEO.py:1247: DeprecationWarning: invalid escape sequence \d
  """
Running tests at all levels
Running zope.testrunner.layer.UnitTests tests:
  Set up zope.testrunner.layer.UnitTests in 0.000 seconds.
  Running:
 testClientBasics (ZEO.asyncio.tests.ClientTests)

Failure in test testClientBasics (ZEO.asyncio.tests.ClientTests)
Traceback (most recent call last):
  File "/usr/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/home/jugmac00/Projects/ZEO/src/ZEO/asyncio/tests.py", line 311, in testClientBasics
    self.assertEqual(self.pop(), (2, False, 'get_info', ()))
  File "/usr/lib/python3.6/unittest/case.py", line 829, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.6/unittest/case.py", line 822, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: [] != (2, False, 'get_info', ())
```

```
Error in test checkQuickVerificationWith2Clients (ZEO.tests.testConnection.FileStorageInvqTests)
Traceback (most recent call last):
  File "/usr/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/home/jugmac00/Projects/ZEO/src/ZEO/tests/ConnectionTests.py", line 598, in checkQuickVerificationWith2Clients
    perstorage = self.openClientStorage(cache="test")
  File "/home/jugmac00/Projects/ZEO/src/ZEO/tests/ConnectionTests.py", line 151, in openClientStorage
    **self._client_options())
  File "/home/jugmac00/Projects/ZEO/src/ZEO/ClientStorage.py", line 281, in __init__
[...]

PR #141, opened 2019-04-25
Use a temporary directory in tests

mgedmin:
Otherwise when you run `tox` this test will create a `data.fs` using Python 2 and then the same test will fail on Python 3 because there already is a `data.fs` in the current working directory, but it has an incompatible format.

See https://github.com/zopefoundation/ZEO/issues/140#issuecomment-486650942

### DIFF
--- a/src/ZEO/tests/test_client_side_conflict_resolution.py
+++ b/src/ZEO/tests/test_client_side_conflict_resolution.py
@@ -1,10 +1,13 @@
+import os
+import shutil
+import tempfile
 import unittest
-import zope.testing.setupstack
 
+import zope.testing.setupstack
 from BTrees.Length import Length
 from ZODB import serialize
 from ZODB.DemoStorage import DemoStorage
-from ZODB.utils import p64, z64, maxtid
+from ZODB.utils import z64, maxtid
 from ZODB.broken import find_global
 
 import ZEO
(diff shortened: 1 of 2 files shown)


src/ZEO/tests/test_client_side_conflict_resolution.py
```diff
@@ -1,10 +1,13 @@
+import os
+import shutil
+import tempfile
 import unittest
-import zope.testing.setupstack
 
+import zope.testing.setupstack
```
jamadden: I don't see this new import being used?

mgedmin: I moved it around, it's not actually new.

This module defines a base class for the test case classes with some helpers (that became  obsolete when stdlib's unittest gained `self.addCleanup()` back in 2.7).

mgedmin:
> Perhaps ironically, one of the Travis builds succumbed to test flakiness and failed. I've restarted it and hopefully it goes green.

It's a shame Travis doesn't let one see the failure log after a job gets restarted. :(

src/ZEO/tests/test_client_side_conflict_resolution.py
jamadden: D'oh. I don't know how I missed the `-` line. Sorry!

jugmac00:
@mgedmin Maybe `Data.fs*` should be removed from .gitignore, too?

https://github.com/zopefoundation/ZEO/blob/a6c1ffcefda76354b07e56af463d5da8e90cf60c/.gitignore#L12

mgedmin:
> Maybe Data.fs* should be removed from .gitignore, too?

`git grep` tells me there are other tests that create `data.fs` files.  I've no idea why those don't cause similar Py2 vs Py3 problems with `tox`...  Anyway, updating .gitignore might be a bit premature.

mgedmin:
> I've no idea why those don't cause similar Py2 vs Py3 problems with tox...

That might be explained by the fact that I was only running _unit_ tests, due to the way tox.ini is structured.  If tests from other layers have similar problems, I've missed them.

PR #145, opened 2019-04-26
fix Py3 SSL tests by providing proper CA test certificates

d-maurer:


d-maurer:
For certificate verification, OpenSSL (and therefore Python 3)
silently ignores certificates without the X509v3 extension "CA:TRUE".
The current test certificates are version 1 certificates (and therefore
lack the version 3 extension). Therefore, their verification fails.

This PR provides test certificates with the "CA" extension.

mgedmin:
> BTW what's the expiration date for these new certificates?

`openssl x509 -noout -text -in src/ZEO/tests/client.pem` says Dec 18 08:58:09 3004 GMT.  I suppose we don't have to worry about expiration :)

(It's more likely the algorithms get deprecated, or some other new extension becomes mandatory.)

> Also it'd be nice to have the commands to generate them 

It's actually documented in a comment at the end of src/ZEO/tests/testssl.py.  I guess my question becomes: did you use that procedure and it's a newer OpenSSL version that decided to add the CA extension?  Or did you use a different procedure and that comment (or the CSR file) needs to be updated to ensure the necessary extensions will be present?

d-maurer:
Marius Gedminas wrote at 2019-4-26 03:56 -0700:
> ...
>It's actually documented in a comment at the end of src/ZEO/tests/testssl.py.  I guess my question becomes: did you use that procedure and it's a newer OpenSSL version that decided to add the CA extension?

No.
I used a script I had developped to generate test certificates
for "dm.xmlsec.binding".

The key to get the extension is having
"-extfile ~/etc/ssl/openssl.cnf -extensions v3_ca"
as additional arguments to the "openssl x509 -req" command
-- with an appropriate "openssl.cnf".

mgedmin:
> The key to get the extension is having
"-extfile ~/etc/ssl/openssl.cnf -extensions v3_ca"
as additional arguments to the "openssl x509 -req" command
-- with an appropriate "openssl.cnf".

Could you update the comment with this information, and also detail what's an "appropriate" openssl.cnf, in case we need to do this again in the future?

jugmac00:
And maybe convert the comment into a module docstring at the top? Then it's easier to find.

d-maurer:
Marius Gedminas wrote at 2019-4-26 23:41 -0700:
>> The key to get the extension is having
>"-extfile ~/etc/ssl/openssl.cnf -extensions v3_ca"
>as additional arguments to the "openssl x509 -req" command
>-- with an appropriate "openssl.cnf".
>
>Could you update the comment with this information

What do you mean with **the** comment? You are replying to **a**
comment with this information.

> and also detail what's an "appropriate" openssl.cnf, in case we need to do this again in the future?

I got my knowledge about SSL certificate generation from
"http://www.imacat.idv.tw/tech/sslcerts.html".

An "appropriate" "openssl.cnf" is a likely part of an OpenSSL installation.
On Ubuntu 18.04, I found one under "/etc/ssl/openssl.cnf".

jugmac00:
@d-maurer @mgedmin referred to this source code comment, which could need an update

https://github.com/zopefoundation/ZEO/blob/ceaadc97110e686203ec727f67d2b1fc3f15ad65/src/ZEO/tests/testssl.py#L372-L380

And what I thought it might be a good idea to turn this source code comment at the bottom in a module docstring so it is easier to find.

d-maurer:
Jürgen Gmach wrote at 2019-4-27 01:37 -0700:
>@d-maurer @mgedmin referred to this source code comment, which could need an update
>
>https://github.com/zopefoundation/ZEO/blob/ceaadc97110e686203ec727f67d2b1fc3f15ad65/src/ZEO/tests/testssl.py#L372-L380

Thank you.

I agree that something should happen to this source comment.
I am not sure what this is.

The current test certificates are valid until 3009.
Should they require recreation, then very likely they must
be recreated in a different way (maybe the key is no longer
sufficiently long, or the encryption algorithms outdated
and need to be replaced, ...). This means that what
we write now in the comment will likely not be correct at
the time of the recreation (as was the case for the current comment).

On the other hand, anyone who wants to use ZEO's SSL feature
will likely need to generate working certificates.
Documentation for this is more important than
for a recreation of the test certificates.
Is a place below "tests" adequate for that user related documentation?
If we put it elsewhere, then the source comment could point there.

mgedmin:
> An "appropriate" "openssl.cnf" is a likely part of an OpenSSL installation.
On Ubuntu 18.04, I found one under "/etc/ssl/openssl.cnf".

Ah, I misunderstood and for some reason thought that you had to make a custom openssl.cnf with some tweaks in it.  Thank you for clarifying!

> This means that what
we write now in the comment will likely not be correct at
the time of the recreation (as was the case for the current comment).

That is a good point.

PR #153, opened 2019-12-13
Fix variable name in code example

sblondon:
The variable is named `address` and is called `addr` in the following lines. This PR uses `address` consistently.


icemac:
Thank you.

PR #157, opened 2020-03-14
Documentation draft

azmeuk:
This is an attempt to generate a sphinx documentation.

- `index.rst` has been splitted in several files.
- all the `.txt` and `.rst` files I found were moved to `docs`
- all the `.txt` files were transformed in `.rst`
- documentation can be built with `tox -e docs`

This is not perfect (there are still some display bugs) but still better than having all the pieces of documentation scattered around the repository. I hope we can publish it on readthedocs if it is merged.

(Having *draft* in the PR title might be confusing, I do think it can be merged!)


docs/reference.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. automodule:: ZEO.cache
+    :members:
+    :show-inheritance:
+
+.. automodule:: ZEO.ClientStorage
+    :members:
+    :show-inheritance:
+
+.. automodule:: ZEO._compat
```
jamadden: ZEO doesn't really have a published public API, I think. Most of this should *not* be documented. Probably  `ClientStorage.ClientStorage` is the only thing that should be documented.

docs/protocol.rst
```diff
@@ -1,5 +1,5 @@
 ZEO Network Protocol (sans authentication)
-==========================================
+------------------------------------------
```
jamadden: If changing adornments, please make them more standard, not less standard.

docs/client-cache.rst
```diff
@@ -0,0 +1,49 @@
+ZEO Client Cache
+----------------
```
jamadden: Please use standard adornments for new files.

docs/blob-nfs.rst
```diff
@@ -1,11 +1,10 @@
-===========================================
```
jamadden: I think the adornments were already correct.

docs/reference.rst
azmeuk: I was not sure what needed to be documented or not. What do you think would be the drawbacks to document other parts?

jamadden: I don't see the need to document internals that can, and do, change arbitrarily. Publishing their documentation implies that they are a stable API meant to be used, and that's not true.

azmeuk: I get it. What about `StorageServer`?

jamadden: Here's one (good?) way to determine what should have its API documented: If it's mentioned in the other documentation (i.e., the original README) it should be documented. Otherwise, no.

That document just includes a few things:
- `ZEO.client`
- `ZEO.DB`
- `ZEO.server`
- `ZEO.Exceptions`

azmeuk: Ok I just documented those, plus `ClientStorage` as `ZEO.client` is just a shortcut for `ClientStorage`.

docs/protocol.rst
azmeuk: Done.

docs/client-cache.rst
azmeuk: Done.

docs/blob-nfs.rst
azmeuk: Done.

doc/conf.py
```diff
@@ -0,0 +1,262 @@
...
+
+sys.path.insert(0, os.path.abspath('.'))
+sys.path.insert(0, os.path.abspath(".."))
+
+# -- General configuration -----------------------------------------------------
+
+# If your documentation needs a minimal Sphinx version, state it here.
+#needs_sphinx = '1.0'
+
+version = '5.2.1.dev0'
```
jamadden: To avoid needing to repeat the version (a step that's usually forgotten, so it quickly grows stale) many projects are now using `pkg_resources` to get the version. [zope.interface example](https://github.com/zopefoundation/zope.interface/blob/7d638c3b149222b164794cac3907c32ab19d2655/docs/conf.py#L18).

src/ZEO/__init__.py
```diff
@@ -22,10 +22,16 @@
...
 def client(*args, **kw):
+    """
+    Shortcut for :class:`ZEO.ClientStorage.ClientStorage`.
+    """
     import ZEO.ClientStorage
     return ZEO.ClientStorage.ClientStorage(*args, **kw)
 
 def DB(*args, **kw):
+    """
+    Shortcut for :class:`ZODB.DB`.
```
jamadden: This might mention that it's using `client`.
```suggestion
    Shortcut for creating a :class:`ZODB.DB` using a ZEO :func:`client`.
```

doc/conf.py
azmeuk: Nice!

azmeuk:
This is merged, can you trigger the RTD build?
Thank you for taking time to review this @jamadden

PR #158, opened 2020-03-27
Update package link from pypi.python.org to pypi.org

sblondon:
This PR updates a link to zdaemon package in the documentation.


sblondon:
The failure is about a bad file descriptor in test 'blob_importexport.txt'. It seems not to be related to this PR.

PR #160, opened 2020-07-10
Include both modified and just created objects into invalidations

navytux:
Starting from 1999 (b3805a2f "just getting started") only modified - not
just created - objects were included into ZEO invalidation messages:

https://github.com/zopefoundation/ZEO/commit/b3805a2f#diff-52fb76aaf08a1643cdb8fdaf69e37802R126-R127

In 2000 this behaviour was further changed to not send invalidation
message at all if the only objects a transaction has were the created ones:

https://github.com/zopefoundation/ZEO/commit/230ffbe8#diff-52fb76aaf08a1643cdb8fdaf69e37802L163-R163

In 2016 the latter was reconsidered as bug and fixed in ZEO5 because
ZODB5 relies more heavily on MVCC semantic and needs to be notified
about every transaction committed to storage to be able to properly
update ZODB.Connection view:

https://github.com/zopefoundation/ZEO/commit/02943acd#diff-52fb76aaf08a1643cdb8fdaf69e37802L889-R834
https://github.com/zopefoundation/ZEO/commit/9613f09b

In 2020, with this patch, I'm proposing to reconsider initial "send only
modified, not created objects" as bug, and include both modified and
just created objects into invalidation messages at least for the
following reasons:

- a ZODB client (not necessarily native ZODB/py client) can maintain
  raw cache for the storage. If such client tries to load an oid at
  database view when that object did not existed yet, gets "no object"
  reply and stores that information into raw cache, to properly invalidate
  the cache it needs an invalidation message from ZODB server that
  _includes_ created object.

- tools like `zodb watch` [1,2,3] don't work properly (give incorrect output)
  if not all objects modified/created by a transaction are included into
  invalidation messages.

- similarly to `zodb watch`, a monitoring tool, that would want to be
  notified of all created/modified objects, won't see full
  database-change picture, and so won't work properly without knowing
  which objects were created.

- wendelin.core 2 - which builds data from ZODB BTrees and data objects
  into virtual filesystem - needs to get invalidation messages with both
  modified and created objects to properly implement its own lazy
  invalidation and isolation protocol for file blocks in OS cache: when
  a block of file is accessed, all clients, that have this block mmaped,
  need to be notified and asked to remmap that block into particular
  revision of the file depending on a client's view of the filesystem and
  database [4,5].

  To compute to where a client needs to remmap the block, WCFS server
  (that in turn acts as ZODB client wrt ZEO/NEO server), needs to be able
  to see whether client's view of the filesystem is before object creation
[...]

PR #161, opened 2020-07-13
[ZEO4] Include both modified and just created objects into invalidations

navytux:
This is ZEO4 backport of https://github.com/zopefoundation/ZEO/pull/160.

Please see details in the first patch.
I also had to backport 8d7b1ceb to get tests back to passing(*) becuase of previously hidden errror. Please see details in the second patch.

Thanks beforehand,
Kirill

/cc @jimfulton 

(*) to the same state as on current 4 branch.

### DIFF
--- a/src/ZEO/StorageServer.py
+++ b/src/ZEO/StorageServer.py
@@ -406,6 +406,7 @@ class ZEOStorage:
 
         self.stats.commits += 1
         self.storage.tpc_finish(self.transaction, self._invalidate)
+        self.client.info(self.get_size_info())
         # Note that the tid is still current because we still hold the
         # commit lock. We'll relinquish it in _clear_transaction.
         tid = self.storage.lastTransaction()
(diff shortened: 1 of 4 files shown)


navytux:
@jamadden, thanks for feedback. I can understand you. On ther other hand: we are still using ZEO4/ZODB4 in production, so for us this kind of change is necessary. It is needed not only for wendelin.core 2, but also for implementing `zconn_at` for ZODB4 which cannot afford to skip being notified of "create-only" transaction. If 4 branch is "dead" we should probably indeed have our fork for custom 4-specific changes until ZODB5 migration is done. I thought this change is useful and tried to go upstream first...

/cc @gidzit, @klawlf82, @jwolf083

navytux:
As suggested I'm moving ZEO4 patches to our fork: https://lab.nexedi.com/kirr/ZEO/commits/4-nxd .

PR #164, opened 2020-12-11
Fix typo client_label -> client-label for strings configuration files

dwt:
I suspect the other underscore containing settings have the same typo - but I hadn't have time to check it, so I'm leaving it as a suggestion for further work. Sorry 'bout that.

### DIFF
--- a/doc/clients.rst
+++ b/doc/clients.rst
@@ -219,7 +219,7 @@ read-only-fallback
 server-sync
    Sets the ``server_sync`` option described above.
 
-wait_timeout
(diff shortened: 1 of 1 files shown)


jamadden (approved):

mgedmin (approved):
LGTM.

And I think you're right, and at least `wait_timeout` right above this should be `wait-timeout`.

dataflake:
I fixed `wait_timeout` to use the dash and removed `disconnect_poll`, which is not configurable in a configuration file.

Issue #155, opened 2020-01-21
Data corruption due to race condition with external invalidations (ZEO5 regression)


jmuchemb:
I submitted zopefoundation/ZODB#291 to fix a similar issue (zopefoundation/ZODB#290) but one of the scripts written by @navytux still fails with ZEO5, ~~when each worker thread has its own connection to the ZEO server~~. I was ~~also~~ unable to reproduce this issue with NEO, so I think it is a different bug, which is in ZEO5.

The script is https://gist.github.com/jmuchemb/b082dadd28c8a6a55c68e510b49873fd

@jamadden Can you try it with RelStorage ?

jamadden:
I ran the script with two different RelStorage backends and didn't see the `AssertionError` after several minutes. I also tried gevent instead of native threads, because that can change timings, and also saw no AssertionError.

jmuchemb:
(I edited the issue description because it's not only about Connection.open())

I think there's a race condition between Client.loadBefore and Client.invalidateTransaction. That's what the cache trace shows:
```
                   loads    hits  inv(h)  writes hitrate
Jan 21 20:26:41 00 '' 0000000000000000 0000000000000000 -
Jan 21 20:26-26       0       0       0       0     n/a
Jan 21 20:26:41 ==================== Restart ====================
Jan 21 20:26:41 20                0 0000000000000000 0000000000000000 -
Jan 21 20:26:41 24                0 0000000000000000 03d5510eaf3e4145 -
Jan 21 20:26:41 52                0 03d5510eaf3e4144 0000000000000000 - 122
Jan 21 20:26:41 22                0 03d5510eaf3e4144 0000000000000000 - 122
Jan 21 20:26:41 22                0 03d5510eaf3e4144 0000000000000000 - 122
# Load obj1
Jan 21 20:26:41 20                1 0000000000000000 0000000000000000 -
Jan 21 20:26:41 24                1 0000000000000000 03d5510eaf3e4145 -
Jan 21 20:26:41 52                1 03d5510eaf3e4144 0000000000000000 - 32
# Load obj2
Jan 21 20:26:41 20                2 0000000000000000 0000000000000000 -
Jan 21 20:26:41 24                2 0000000000000000 03d5510eaf3e4145 -
# Waiting from obj2 to be loaded from server.
# Process invalidations from a recently committed transactions.
Jan 21 20:26:41 1c                1 03d5510eaf7847cc 0000000000000000 -
[...]

PR #167, opened 2021-04-08
move `setLastTID` call after connection cache invalidation processing

### LINKED ISSUE #166: Another data corruption due to concurrency bug  (ZEO5 regression; no external invalidations)
( moving out from https://github.com/zopefoundation/ZEO/issues/155#issuecomment-811984478 since this seems to be a different problem )

Something really wrong is going on with ZEO5.

In addition to the test program provided in https://github.com/zopefoundation/ZEO/issues/155, where there are 8 simultaneous workers each of which _opens its own storage-level connection_, I've tried to run the test program from https://github.com/zopefoundation/ZODB/issues/290 with minimal edit to use ZEO instead of `MappingStorage`, and it **fails too**.

In that test program **there is only one storage-level connection** and two simultaneous users that are using **that single** storage-level connection. In other words there are **no external invalidations** involved at all.

https://lab.nexedi.com/kirr/wendelin.core/blob/e2dc8651/lib/tests/testprog/zopenrace.py#L102-203

This failure indicates **data corruption**. Quoting https://github.com/zopefoundation/ZODB/issues/290:

---- 8< ----
That in turn can lead to loose of Consistency of the database if mix of current
and old data is used to process a transaction. A classic example would be bank
accounts A, B and C with A<-B and A<-C transfer transactions. If transaction
that handles A<-C sees stale data for A when starting its processing, it
results in A loosing what it should have received from B.

...

The program simulates two clients: one (T2) constantly modifies two integer
objects preserving invariant that their values stay equal. The other client
(T1) constantly opens the database and verifies the invariant. T1 forces access
to one of the object to always go through loading from the database, and this
way if live cache becomes stale the bug is observed as invariant breakage.
---- 8< ----

Here is how it fails with ZEO:

```
.../wendelin.core$ WENDELIN_CORE_TEST_DB="<zeo>" py.test -vs -k openrace lib/tests/test_zodb.py
============================== test session starts ==============================
platform linux2 -- Python 2.7.18, pytest-4.6.11, py-1.9.0, pluggy-0.13.1 -- /home/kirr/src/wendelin/venv/z-dev/bin/python2
cachedir: .pytest_cache
rootdir: /home/kirr/src/wendelin/wendelin.core
collected 8 items / 6 deselected / 2 selected

lib/tests/test_zodb.py::test_zodb_zopenrace_basic OK
PASSED
lib/tests/test_zodb.py::test_zodb_zopenrace FAILED

=================================== FAILURES ====================================
______________________________ test_zodb_zopenrace ______________________________

    def test_zodb_zopenrace():
        # exercises ZODB.Connection + particular storage implementation
>       zopenrace.main()

lib/tests/test_zodb.py:382:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
<decorator-gen-2>:2: in main
    ???
../../tools/go/pygolang/golang/__init__.py:103: in _
    return f(*argv, **kw)
lib/tests/testprog/zopenrace.py:115: in main
    test(zstor)
<decorator-gen-3>:2: in test
    ???
../../tools/go/pygolang/golang/__init__.py:103: in _
    return f(*argv, **kw)
lib/tests/testprog/zopenrace.py:201: in test
    wg.wait()
golang/_sync.pyx:246: in golang._sync.PyWorkGroup.wait
    ???
golang/_sync.pyx:226: in golang._sync.PyWorkGroup.go.pyrunf
    ???
lib/tests/testprog/zopenrace.py:165: in T1
    t1()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def t1():
        transaction.begin()
        zconn = db.open()

        root = zconn.root()
        obj1 = root['obj1']
        obj2 = root['obj2']

        # obj1 - reload it from zstor
        # obj2 - get it from zconn cache
        obj1._p_invalidate()

        # both objects must have the same values
        i1 = obj1.i
        i2 = obj2.i
        if i1 != i2:
>           raise AssertionError("T1: obj1.i (%d)  !=  obj2.i (%d)" % (i1, i2))
E           AssertionError: T1: obj1.i (16)  !=  obj2.i (15)

lib/tests/testprog/zopenrace.py:156: AssertionError
```

P.S. Suggested patch from https://github.com/zopefoundation/ZEO/issues/155#issuecomment-581046248 does not help here - even though it seemingly fixes https://github.com/zopefoundation/ZEO/issues/155, it does not help to fix https://github.com/zopefoundation/ZODB/issues/290 with ZEO5.

P.P.S. ZEO `5.2.2-1-g3d90ed42`, ZODB `5.6.0-18-g75c703fcf`, CPython `2.7.18`

d-maurer:
… as required by ZODB>=5.6.0

This fixes #166.

The PR no longer actually moves the call but instead uses locking to ensure that foreign threads see the effect of `setLastTid` only after the callback call.


jmuchemb:
I haven't reviewed the new test but this PR looks enough.

> Not sure what happens with ZODB<5.6.0.

I'd say we don't care. Using ZODB<5.6.0 is already a risk to have data corruption.

d-maurer:
The PR now depends on "https://github.com/zopefoundation/ZODB/pull/344" to allow both `RaceConditionTests.test_open` as well as `ZODB.tests.BasicStorage.BasicStorage.check_tid_ordering_w_commit` to succeed.

2 tests are still failing:

    1. ClientStorageNonSharedBlobsBlobTests:commit_from_wrong_partition (ZODB.tests.testblob)
    2. ClientStorageSharedBlobsBlobTests:commit_from_wrong_partition (ZODB.tests.testblob)
 
Those failures are independent from the PR (they fail also for `master`). They are doctests and use logging to verify that things happen as they should. This may not work in a ZEO setup.

src/ZEO/asyncio/client.py
```diff
@@ -657,7 +655,26 @@ def tpc_finish_threadsafe(self, future, wait_ready, tid, updates, f):
...
+                    # This is the case for "normal" operations involving
+                    # the ``MVCCAdapter``.
+                    # From ``ZODB>=5.6`` on, ``MVCCAdapter`` requires
+                    # that `lastTransaction` changes after invalidation
+                    # processing; therefore, `setLastTid` is called
+                    # after the callback
+                    f(tid)
+                    cache.setLastTid(tid)
+                else:  # called from tests
+                    # the call below would deadlock for older ``ZODB``
```
navytux, line 669: @d-maurer, while I appreciate your work on getting ZEO5 fixed, to me going via this ad-hoc way would not be good. Like @jmuchemb said ZODB < 5.6 is already at risk of data corruption (see https://github.com/zopefoundation/ZODB/issues/290). From this point of view maybe more logical would be to just make ZEO5 require ZODB>=5.6 ?

I have also left some comments regarding `check_tid_ordering_w_commit` logic in https://github.com/zopefoundation/ZODB/pull/307#issuecomment-818926431 .

d-maurer:
Kirill Smelkov wrote at 2021-4-13 10:49 -0700:
***@***.*** commented on this pull request.
> ...
***@***.***, while I appreciate your work on getting ZEO5 fixed, to me going via this ad-hoc way would not be good. Like @jmuchemb said ZODB < 5.6 is already at risk of data corruption (see https://github.com/zopefoundation/ZODB/issues/290). From this point of view maybe more logical would be to just make ZEO5 require ZODB>=5.6 ?
>
>I have also left some comments regarding `check_tid_ordering_w_commit` logic in https://github.com/zopefoundation/ZODB/pull/307#issuecomment-818926431 .

Based on that comment, I think that `check_tid_ordering_w_commit`
fails for ZEO5 because its `ClientStorage` does not implement
the storage lock. The right way is likely to employ some lock
to get a similar effect.

@jmuchemb earlier proposed to use the cache's lock for this purpose.
Due to my misunderstanding of the above test, I failed to recognize
its potential. Now, I think that this is the best approach.

d-maurer:
The failing tests are unrelated to this PR.

dataflake (commented):
I don't know enough about ZEO or the underlying ZODB requirements to review this change.

navytux:
@d-maurer, I've tried to review your updated fix. I have limited understanding
of what is going on in ZEO5 and asyncio, so I might be missing something. Still
maybe my feedback might be useful.

First, let's note that in ZEO3 and ZEO4 invalidations were processed in atomic way with
utilising dedicated `ClientStorage._lock`. See c26e6cd7 for where this atomic
processing originated, and also the following code snippets for latest ZEO4:

https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L376-L380
https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L830-L852
https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L1121-L1123
https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L1159-L1167
https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L1459-L1471

That atomic invalidation processing was lost during ZEO5 development, and in
essence we are trying to reintroduce it.

The protection could be done with bringing back `ClientStorage._lock`, or via
utilising `ClientStorage._cache._lock` - like you do in this patch - because
`_cache.lock` is recursive. Personally I find using dedicated lock to be more
clear, but using `._cache._lock` for this unintended purpose could also work in practice.

When comparing ZEO5 state to what was in ZEO4, even with this patch, I can see
at least the following differences:

- `invalidateTransaction` still processes invalidations in non-atomic way. Suggested amendment:

```diff
@@ -679,10 +681,11 @@ def close_threadsafe(self, future, _):
 
     def invalidateTransaction(self, tid, oids):
         if self.ready:
-            for oid in oids:
-                self.cache.invalidate(oid, tid)
-            self.client.invalidateTransaction(tid, oids)
-            self.cache.setLastTid(tid)
+            with self.cache._lock: # atomic processing of invalidations
+                for oid in oids:
+                    self.cache.invalidate(oid, tid)
+                self.client.invalidateTransaction(tid, oids)
+                self.cache.setLastTid(tid)
         else:
             self.verify_invalidation_queue.append((tid, oids))
```

- ZEO4 client was calling `_server.tpc_finish` from under `._lock`:

  https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L1159-L1167

  This behaviour was introduced in a83d0f32 which talks about `"A ZEO threading bug could cause transactions to read inconsistent data"` so it might be important to analyze because in ZEO5 `._server.tpc_finish` is called and waited-for without holding any lock.

When analyzing this latter topic let's first note that it is generally _unsafe_ to use
regular locks from under asyncio. It is generally unsafe because if taking such
regular lock blocks, then event loop will be stuck waiting for the lock
acquisition, and so all other queued asynchronous jobs will pause to be processed
until then. Now if another thread queues asynchronous task, and waits for its
completion with holding such regular lock, it will cause deadlock: the thread is
waiting for async task to complete, the lock is held. Asyncio loop reschedules
another async task which tries to acquire the lock -> deadlock. That's why for
asyncio they also provide [asyncio-specific
Lock](https://docs.python.org/3/library/asyncio-sync.html) and other
synchronization primitives.

Now, we are already using _regular_ `._cache._lock` from under asyncio
`ClientThread`. After going through ZEO5 codebase, it seems, this usage is
_safe_, because nothing waits for async task completion with holding
that `._cache._lock`.

BUT we cannot reintroduce ZEO4 behaviour and change `ClientStorage.tpc_finish`
to call `._server.tpc_finish` with holding "atomic invalidations" lock, because
it will create the exact asyncio/regular-lock deadlock scenario explained above.

I _think_ in ZEO5 it is safe to call `._server.tpc_finish` and wait for its
completion without holding the lock and acquire it only inside tpc_finish
handler when processing invalidations. My thinking here is that:

- in `ClientStorage.tpc_finish` we send to the server request to finish a transaction
- before we get a reply the server might send us several `invalidateTransaction` messages.
  With my amendment we handle that invalidations atomically.
- the server will eventually send us reply for tpc_finish message.

  The server guarantees ordering for invalidateTransaction and tpc_finish reply:
  For all those messages corresponding `.tid` is monotonically increasing.
  In other words we cannot get tpc_finish reply with tid1, and later
  invalidateTransaction with tid2<tid1 or vise-versa. This way even while we are waiting for
  the reply the cache is kept consistent even with it's `._ltid` potentially
  growing during the wait.

  in a sense tpc_finish reply handler is very similar to handler of
  invalidateTransaction except that it also stores to the cache just committed
  data and calls user-provided callback.

  And tpc_finish callback is documented in ZODB interfaces to be called with holding storage transaction lock:

  https://github.com/zopefoundation/ZODB/blob/dad778016f22f45a625a6914b21017e9af3bb03d/src/ZODB/interfaces.py#L852-L854

  so it cannot do arbitrary storage calls.

  In practice this callback is used only in MVCCAdapter to update local idea of
  last tid and what was invalidated.

- even if we have several tpc_finish called simultaneously from one client, there handlers are still run one-by-one and updates to the cache and invalidations are done in atomic way.

So in the end the core of your patch looks correct.

I would suggest to consider adding a note that `with cache._lock` in tpc_finish
handler is used for "atomic invalidation processing", something like

```diff
@@ -655,6 +655,8 @@ def tpc_finish_threadsafe(self, future, wait_ready, tid, updates, f):
                     # the effect of ``setLastTid`` becomes visible
                     # by other threads only after the ``f`` callback
                     # has run as required by ``ZODB>=5.6``.
+                    #
+                    # it is also required to process invalidations atomically.
                     for oid, data, resolved in updates:
                         cache.invalidate(oid, tid)
                         if data and not resolved:
```

Also, if we are adding `._lock` to `MemoryCache` it is strange that we are not
using it inside the `MemoryCache` itself. With introducing
`ClientStorage._lock` with the intended application to process invalidation
atomically, this inconsistency will automatically go away.


Hope it helps a bit,
Kirill


P.S. And I've moved the test to ZODB itself so that it is exercised wrt all storages, not
only for ZEO: https://github.com/zopefoundation/ZODB/pull/345.

d-maurer:
Kirill Smelkov wrote at 2021-4-14 07:57 -0700:
> ...
>That atomic invalidation processing was lost during ZEO5 development, and in
>essence we are trying to reintroduce it.
>
>The protection could be done with bringing back `ClientStorage._lock`, or via
>utilising `ClientStorage._cache._lock` - like you do in this patch - because
>`_cache.lock` is recursive. Personally I find using dedicated lock to be more
>clear, but using `._cache._lock` for this unintended purpose could also work in practice.

Julien proposed `cache._lock` because it is this lock that (effectively)
protects `lastTransaction` and `lastTransaction has been
responsible for the problem in #166.


>When comparing ZEO5 state to what was in ZEO4, even with this patch, I can see
>at least the following differences:
>
>- `invalidateTransaction` still processes invalidations in non-atomic way. Suggested amendment:
>
>```diff
>@@ -679,10 +681,11 @@ def close_threadsafe(self, future, _):
>
>     def invalidateTransaction(self, tid, oids):
>         if self.ready:
>-            for oid in oids:
>-                self.cache.invalidate(oid, tid)
>-            self.client.invalidateTransaction(tid, oids)
>-            self.cache.setLastTid(tid)
>+            with self.cache._lock: # atomic processing of invalidations
>+                for oid in oids:
>+                    self.cache.invalidate(oid, tid)
>+                self.client.invalidateTransaction(tid, oids)
>+                self.cache.setLastTid(tid)
>         else:
>             self.verify_invalidation_queue.append((tid, oids))
>```

I think locking is not necessary here:

 * `invalidateTransaction` and all functions invoked by it
   are only called by coroutine code
   The coroutine code is essentially sequential with
   potential (coroutine) context switches only where
   futures are yielded.
   Because `invalidateTransaction contains no `yield`
   there is no need to protect it against other coroutine code.

 * unlike the situation in `tpc_finish_threadsafe`
   `setLastTid` is called after invalidation processing.
   We do not need a lock to ensure that concurrent threads
   can observe the `lastTransaction` change only
   after invalidation processing.

> ...
>  https://github.com/zopefoundation/ZEO/blob/47d3fbe8cbf24cad91b183483df069ef20708874/src/ZEO/ClientStorage.py#L1159-L1167
>
>  This behaviour was introduced in a83d0f32 which talks about `"A ZEO threading bug could cause transactions to read inconsistent data"` so it might be important to analyze because in ZEO5 `._server.tpc_finish` is called and waited-for without holding any lock.

My comment in `tpc_finish_threadsafe` should address this question.
Because this function is not protected by a lock
there is the potential of a race condition: there is
a time span where the server delivers already new data while
the cache is not yet updated. This race can affect "load current"
operations. However, the "normal" load does not perform
"load current" but "load at transaction start time"
where the effective "transaction start time" is determined
via a call of `lastTransaction` (and therefore depends on
`cache.setLastTid`). This means that for "normal" loads
inconsistencies for transactions later than `lastTransaction`
are not relevant.

> ...
>Now, we are already using _regular_ `._cache._lock` from under asyncio
>`ClientThread`. After going through ZEO5 codebase, it seems, this usage is
>_safe_, because nothing waits for async task completion with holding
>that `._cache._lock`.

Right.

> ...
>Also, if we are adding `._lock` to `MemoryCache` it is strange that we are not
>using it inside the `MemoryCache` itself. With introducing
>`ClientStorage._lock` with the intended application to process invalidation
>atomically, this inconsistency will automatically go away.

`MemoryCache` is used only for tests -- as a simple mockup
of a true client cache. Mockup emulations do not need to be faithful.

>Hope it helps a bit,

Thank you very much!
I highly appreciate our discussion.

> ...
>P.S. And I've moved the test to ZODB itself so that it is exercised wrt all storages, not
>only for ZEO: https://github.com/zopefoundation/ZODB/pull/345.

Good!

We might use this to convert the ZEO test into a true
multi (2) client test (as sketched above).

d-maurer:
Kirill Smelkov wrote at 2021-4-14 11:41 -0700:
> ...
>If it were only coroutine code then you would be right. And still it would be
>very implicit and deserving at least a comment. However the locking is needed
>for protecting "coroutine code" wrt other _regular_ threads. Please look, for
>example, at `ClientStorage.loadBefore`:
>
>https://github.com/zopefoundation/ZEO/blob/3d90ed421b22b403ae32cf5d76ced53409479ecb/src/ZEO/ClientStorage.py#L519-L524
>
>This code is called simultaneously by potentially several _regular_ threads.
>
>Do we want to keep on reasoning about that that code is correct if it meets the
>cache that was only half-way invalidated by asyncio?

I do not know how much **we** want to reason.

Looking at the (ZEO5) code, I got the impression that it tries
to reduce locking to a minumum -- e.g. a comment expresses
sadness that the commit lock could not yet get eliminated.
It would be in line to acquire locks only where necessary.

As you have pointed out locking in asynchronous code can easily
lead to deadlocks.

MVCC implies that regular loads are always for tids <= `lastTransaction`.
Therefore, invalidations for tids > `lastTransaction` should not matter
-- even if they are not yet fully performed.

>Even if it can be proved
>correct, what are the upsides of not having a big "atomic invalidations" lock
>around all that update? For ZEO5 there is already *two* **data corruption**
>bugs that originated from ZEO4->ZEO5 rework in this area. I would say: just be
>on the safe side and keep things simple and robust. _If_ having that lock turns
>out to be too expensive to have, then it is `cache._lock` that is better to
>remove and keep a bigger `with ClientStorage._lock` around all cache access.

One drawback of ZEO has been its relative slowness partially caused by its use
of quite coarse grained locks. I assume that lock usage was reduced
to increase ZEO speed. I would try to avoid locks spanning server
interactions (of course, the lock around invalidation processing would
not include such a span).


>In fact ClientCache._lock can be dropped completely if all access to the cache
>is protected with ClientStorage._lock. Then even performance-wise it will be a
>bit faster compared to what we currently have. And automatically more
>robust/clear and correct by design.

Sure. But the authors had something in mind by moving locking
from `ClientStorage` to the cache. One of those reasons might
be that cache locking tends to be finer grained than storage locking.

> ...
>> We might use this to convert the ZEO test into a true
>> multi (2) client test (as sketched above).
>
>I'm not sure I understand. Could you please clarify this part?

I think we currently do not have a race condition test
for multiple client storages. The test you provided for #166
only examines race conditions associated with a single client storage,
caused by `tpc_finish` races.

Above, I proposed to modify that test to use 2 client storages
(rather than a single one) and probe for race conditions
associated with asynchronous calls of `invalidateTransaction`.

jmuchemb:
> @jmuchemb earlier proposed to use the cache's lock for this purpose.

Not exactly. It was only a try to argue about changing ZODB or not. I also wrote that we should ideally prove that changing how ZEO locks is necessary, because maybe it's simply the `check_tid_ordering_w_commit` test that is just too strict.

About NEO:
> when the load is processed, the last_tid is guaranteed to be updated (NEO needed that with ZODB<5).

I means that when NEO drops support for ZODB<5, I may decide to update last_tid after that the lock is released, simply to reduce the probability of lock contention (but this update is such small code that it's probably not worth it). Then I'd need to make `check_tid_ordering_w_commit` less strict again: I tried this change in NEO and the test fails occasionally.

Back to ZEO:
> Do we want to keep on reasoning about that the code is correct if it meets the
> cache that was only half-way invalidated by asyncio? Even if it can be proved
> correct, what are the upsides of not having a big "atomic invalidations" lock
> around all that update? For ZEO5 there is already _two_ **data corruption**
> bugs that originated from ZEO4->ZEO5 rework in this area.

I don't think it's so difficult to prove that we don't need to extra locking. But I'm not familiar enough with ZEO code to do it now without spending much time, and anyway...

> I would say: just be
> on the safe side and keep things simple and robust. _If_ having that lock turns
> out to be too expensive to have, then it is `cache._lock` that is better to
> remove and keep a bigger `with ClientStorage._lock` around all cache access.

yes, we should ideally do that. Besides the ugliness of accessing `_lock` from outside `ClientCache`, a lock that is internal to the cache is expensive because it often needs to be taken many times in a row, and `RLock` is slower than `Lock`.

d-maurer:
Julien Muchembled wrote at 2021-4-14 16:39 -0700:
>> @jmuchemb earlier proposed to use the cache's lock for this purpose.
>
>Not exactly. It was only a try to argue about changing ZODB or not. I also wrote that we should ideally prove that changing how ZEO locks is necessary, because maybe it's simply the `check_tid_ordering_w_commit` test that is just too strict.

Since ZODB 5.6 it is vital that `lastTransaction` changes its return
value at a precise moment. A strict test should verify this
requirement (however, this test is not necessarily `check_tid_ordering...`).

d-maurer:
Kirill Smelkov wrote at 2021-4-14 11:41 -0700:
> ...
>> We might use this to convert the ZEO test into a true
>> multi (2) client test (as sketched above).
>
>I'm not sure I understand. Could you please clarify this part?

I have added such a test
(--> `ZEO.tests.testZEO.RaceConditionTests.test_race_two_storages`).
It is almost your test developed for #166 but uses two
databases (and therefore storages), one for mutations and one for checks.
It therefore monitors potential race conditions associated
with asynchronous `invalidateTransaction` calls
(rather than invalidations inside `tpc_finish`).

navytux:
@d-maurer, @jmuchemb, thanks for feedback.

I understand that during ZEO5 development the intention was to reduce
unnecessary locking overhead. But I also feel that this transition was not
carefully implemented, and also that original author somewhat fall into usual
asyncio trap, thinking that "we are free from the races because we do only one
thing at a time", while in reality asyncio provides only protection from "data
races", not from general race conditions.

The fact that during ZODB4/ZEO4->ZODB5/ZEO5 development several bugs in
concurrency area with data-corruption consequences were introduced justify my
words. We are, after all, trying to fix one of those regressions here.

I also agree that it is not difficult to prove that lack of overall locking in
`invalidateTransaction` does not lead to any bad behaviour (at least it _seems_
so to me). However I also have an evidence that concurrency bugs are too often
hide under seemingly correct constructs that people tend to argue about with
saying something like "this is correct - no need to change anything". To
justify my experience let me show one example of concurrency bug in Python
implementation of `PyThread_release_lock` that was leading to memory corruption
and deadlocks. Quoting https://bugs.python.org/issue38106:

>    Bug history
>
>    The bug was there since 1994 - since at least [1]. It was discussed in 2001
>    with original code author[2], but the code was still considered to be
>    race-free. In 2010 the place where pthread_cond_signal should be - before or
>    after pthread_mutex_unlock - was discussed with the rationale to avoid
>    threads bouncing[3,4,5], and in 2012 pthread_cond_signal was moved to be
>    called from under mutex, but only for CPython3[6,7].
>    
>    In 2019 the bug was (re-)discovered while testing Pygolang[8] on macOS with
>    CPython2 and PyPy2 and PyPy3.
>    
>    [1] https://github.com/python/cpython/commit/2c8cb9f3d240
>    [2] https://bugs.python.org/issue433625
>    [3] https://bugs.python.org/issue8299#msg103224
>    [4] https://bugs.python.org/issue8410#msg103313
>    [5] https://bugs.python.org/issue8411#msg113301
>    [6] https://bugs.python.org/issue15038#msg163187
>    [7] https://github.com/python/cpython/commit/187aa545165d
>    [8] https://pypi.org/project/pygolang

So it took 25 years to find and fix, even though in 2001 there was a discussion
and proposal to (correctly) rework the code. However at that time the proposal
was rejected because original code author was believing it to be correct (see
[2] above). After that discussion it took another 18 years (!) to go through
rediscover, dig, analyze and fix. And all that could be avoided if people were
following _"just be on the safe side"_ approach.

This way I highly recommend to restore "atomic processing of invalidations" in
`invalidateTransaction` too.

For the reference, it is indeed much better to make a tpc_finish roundtrip
without holding any lock, and I completely agree with this part of ZEO4->ZEO5 change.

--------

Thanks for clarifying about `RaceConditionTests.test_race_two_storages`.

I have added corresponding test for "load vs external invalidation" to ZODB testsuite
(https://github.com/zopefoundation/ZODB/pull/345#issuecomment-821316526). Unfortunately, even though there are several CI failures caused by
`RaceConditionTests.test_race_two_storages` added in this PR, I _cannot_ reproduce
those failures on my computer at all. I've also first tried to implement "load
vs external invalidation" test in ZODB following 1 mutator -> 1 verifier
scheme, but it also does not expose the bug.

This way the test that I've ended up adding into ZODB is based on my original "8
workers modifying/checking the database simultaneously" approach from
https://github.com/zopefoundation/ZEO/issues/155#issuecomment-577602842. With
that approach the probability to hit the bug is very high for me, and at a
small iteration number.

Kirill

d-maurer:
Kirill Smelkov wrote at 2021-4-16 10:18 -0700:
***@***.***, @jmuchemb, thanks for feedback.
> ...
>Unfortunately, even though there are several CI failures caused by
>`RaceConditionTests.test_race_two_storages` added in this PR, I _cannot_ reproduce
>those failures on my computer at all.

Where have you seen those CI failures? When I pushed the test,
there was no failure related to it new test (as I had expected).

navytux:
Here are those failures:

https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673488
https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673489
https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673490
https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673491
https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673498

(search for "Failure in test test_race_two_storages")

I discovered them all by clicking on "[Details](https://github.com/zopefoundation/ZEO/pull/167/checks?check_run_id=2349867794)" right to "Travis CI" status.

d-maurer:
Kirill Smelkov wrote at 2021-4-16 10:58 -0700:
>Here are those failures:
>
>https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673488
>https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673489
>https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673490
>https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673491
>https://travis-ci.com/github/zopefoundation/ZEO/jobs/498673498
>
>(search for "Failure in test test_race_two_storages")
>
>I discovered them all by clicking on "[Details](https://github.com/zopefoundation/ZEO/pull/167/checks?check_run_id=2349867794)" right to "Travis CI" status.

Thank you!

Apparently, I missed these errors.
Especially strange that in one failure "i1" (the value freshly read
from the ZODB) was smaller than "i2" (the potentially cached value).

I am trying to understand why Python 3 is so slow in
my local environment (this can apparently drastically reduce the
likelyhood of race conditions). After that, I will try to find the
cause for the failing `test_race_two_storages`.

navytux:
@d-maurer, you are welcome and good luck with reaching understanding of what is going on with i1 < i2 and Python3. Regarding `test_race_two_storages`, personally, I think that issues are due to https://github.com/zopefoundation/ZEO/issues/155 (and so should be fixed by https://github.com/zopefoundation/ZEO/pull/169), but, of course, I might be missing something.

d-maurer:
@navytux 
> Regarding `test_race_two_storages`, personally, I think that issues are due to #155 (and so should be fixed by #169), but, of course, I might be missing something.

I have merged your #169 and the tests now pass on Travis (as you have expected).

I have also resolved the Travis `uvloop` problem (unrelated to our changes): it is sufficient to update `.travis.yml` for the new `uvloop` version (with requires Python 3.7 or later).

navytux:
@d-maurer, glad to hear that.

navytux:
@d-maurer, do you need any help from me to recover this pull-request from conflicts?

For the reference - I'm generally ok with your changes. I just highly recommend to be on the safe side and add atomicity into processing invalidations in `invalidateTransaction`. But it is still a recommendation and is not obligatory.

navytux (approved):

d-maurer:
Kirill Smelkov wrote at 2021-4-21 23:09 -0700:
***@***.***, do you need any help from me to recover this pull-request from conflicts?

Thank you for your help offer.
Likely, I can cope. If not I will ask for help.

>For the reference - I'm generally ok with your changes. I just highly recommend to be on the safe side and add atomicity into processing invalidations in `invalidateTransaction`. But it is still a recommendation and is not obligatory.

I was convinced that this were not necessary for normal (i.e. MVCC)
load operations. Thinking again, I suppose that there may be
another potential race condition. If this is the case, then
ensuring atomicity alone may not be enough.
Following is a sketch of the potential race condition scenario
which hopefully allows others to think about this case:

Assume we are performing a local `tpc_finish` and during this
time we receive an `invalidateTransaction` from the server
(which necessarily comes from a different transaction
as the server does not sent `invalidateTransaction` for local
commits). The *tid* in the `invalidateTransaction` might be larger
or smaller than our local *tid* but the cache *tid* must be strictly
increasing. Depending on the execution order in `tpc_finish`
and `invalidateTransaction`, the last condition might be violated.

I am not sure that the case can actually happen. If it can
happen, then potentially we can prove that the *tid* in
`invalidateTransaction` is larger than that in the local `tpc_finish`
(because an `invalidateTransaction` with a smaller *tid* would
necessarily have been performed (inside the `async` thread) before the
(asynchronous) `tpc_finish` server call returns).

I will still think some time about the possibility of this case
and potentially make changes to the PR.


I also plan to remove my ZEO race condition tests because
comparable tests now come via ZODB tests
(run as part of the ZEO test suite).
It would help me to have a ZODB release containing those tests:
while I can simply run the ZEO tests locally with any ZODB
branch I would need some time to achieve this for the Travis tests.
And I have seen several cases where race condition tests passed
locally but not on Travis.

navytux:
@d-maurer, thanks for feedback and sharing your thoughs. I've looked around a bit and from my limited understanding I believe that race scenario should not be possible:

- ZEO server takes storage-global "commit lock" on vote:

https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L223
https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L369-L419

- this commit lock is released after tpc_finish completes:

https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L333-L345
https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L357-L359

- this way during local tcp_finish is in progress, the commit lock is already held for our client and no other clients are progressing with their votes.

- transaction ID for committed transaction is assigned by server only at tpc_finish time.

- this way vote+finish for transactions from different clients are processed serially on the server. And server replies with "tpc_finished" and "invalidateTransaction" are also serialized.

In other words the server guarantees serialization for created transactions and serialization for "tpc_finished" and "invalidateTransaction" notification messages. This guarantee includes guarantee for `.tid` being monotonically increased on every such message.

> It would help me to have a ZODB release containing those tests:

I would be glad to help here, but I do not have access rights to make ZODB releases. It is only @dataflake, @jimfulton, @jmuchemb and @tseaver who are currently listed as [ZODB maintainers](https://pypi.org/project/ZODB/) on PyPI. If people give me access I would be glad to try to help.

For the reference: if we are to extend list of ZODB maintainers, I belive @jamadden should be included first. He is already doing (in my view) a good job maintaining ZODB on github, and also he is already in the list of maintainers - and doing a good job there too - for other ZODB-related projects like [BTrees](https://pypi.org/project/BTrees/) and [persistent](https://pypi.org/project/persistent/).

Kirill

d-maurer:
Kirill Smelkov wrote at 2021-4-22 11:03 -0700:
***@***.***, thanks for feedback and sharing your thoughs. I've looked around a bit and from my limited understanding I believe that race scenario should not be possible:
>
>- ZEO server takes storage-global "commit lock" on vote:
>
>https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L223
>https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L369-L419
>
>- this commit lock is released after tpc_finish completes:
>
>https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L333-L345
>https://github.com/zopefoundation/ZEO/blob/e277ba097241ec9a463a3a3b4c79b8c9cc4e7ac9/src/ZEO/StorageServer.py#L357-L359
>
>- this way during local tcp_finish is in progress, the commit lock is already held for our client and no other clients are progressing with their votes.

At this point, I am not sure:
when the client handles cache invalidations, `tpc_finish` has
already terminated for the server (and the commit lock released).
A different client might have performed a new transaction.

>- transaction ID for committed transaction is assigned by server only at tpc_finish time.

That is why I believe that a *tid* reported by `invalidateTransaction`
would necessarily be larger than that from the local `tpc_finish`.

>- this way vote+finish for transactions from different clients are processed serially on the server. And server replies with "tpc_finished" and "invalidateTransaction" are also serialized.

The critical period is the time when then client processes the
results of its `tpc_finish` call: at this time, the `tpc_finish`
transaction is terminated for the server and it may process a new
transaction.

navytux:
@d-maurer, maybe I'm missing something, but the messages that server sent - tpc_finish (with tid1), and follow-up invalidateTransaction (with tid2 > tid1) - will be received in _that_ order on client. And they will also be processed _serially_ in _that_ order on client. This way client cache will be first updated with tid1, and only then with tid2 _exactly_ in this order. Isn't it the case? What I am, maybe, missing?

d-maurer:
Kirill Smelkov wrote at 2021-4-22 19:03 -0700:
***@***.***, maybe I'm missing something, but the messages that server sent - tpc_finish (with tid1), and follow-up invalidateTransaction (with tid2 > tid1) - will be received in _that_ order on client. And they will also be processed _serially_ in _that_ order on client. This way client cache will be first updated with tid1, and only then with tid2 _exactly_ in this order. Isn't it the case? What I am, maybe, missing?

You are right.

I was not sure:
The important point is when (during `tpc_finish`) the switch
between the async thread and the client thread happens.
If the cache is updated in the async thread, then all cache updates
happen in the async thread and are serialized.
Should `tpc_finish` update the cache in the client thread, serialization
would not be garanteed.
I have now verified that the thread switch happens in
`ClientRunner.tpc_finish` and not in `Client.tpc_finish_threadsafe`.
Thus, the cache is updated in the async thread.


There is still one point I have not yet fully understood.
In the discussion about the "loadBefore" race I got the impression
that part of message processing could be out of order.
I think this part refered to the processing of "done callbacks", not
the initial message processing. But, I would like to understand
this better before the merge.

d-maurer:
Replaced by #174 (cleaned up version of this PR).

PR #169, opened 2021-04-14
client: Fix cache corruption on loadBefore and prefetch

navytux:
Currently loadBefore and prefetch spawn async protocol.load_before task,
and, after waking up on its completion, populate the cache with received
data. But in between the time when protocol.load_before handler is run
and the time when protocol.load_before caller wakes up, there is a
window in which event loop might be running some other code, including
the code that handles invalidateTransaction messages from the server.

This means that cache updates and cache invalidations can be processed on
the client not in the order that server sent it. And such difference in
the order can lead to data corruption if e.g server sent

	<- loadBefore oid serial=tid1 next_serial=None
	<- invalidateTransaction tid2 oid

and client processed it as

	invalidateTransaction tid2 oid
	cache.store(oid, tid1, next_serial=None)

because here the end effect is that invalidation for oid@tid2 is not
applied to the cache.

The fix is simple: perform cache updates right after loadBefore reply is
received.

Fixes: https://github.com/zopefoundation/ZEO/issues/155

The fix is based on analysis and initial patch by @jmuchemb:

https://github.com/zopefoundation/ZEO/issues/155#issuecomment-581046248

~~For tests, similarly to https://github.com/zopefoundation/ZODB/pull/345,
I wanted to include a general test for this issue into ZODB, so that all
storages - not just ZEO - are exercised for this race scenario. However
in ZODB test infrastructure there is currently no established general
way to open several client storage connections to one storage server.
This way the test for this issue currently lives in wendelin.core
repository (and exercises both NEO and ZEO there):~~

~~https://lab.nexedi.com/nexedi/wendelin.core/commit/c37a989d~~

~~I understand there is a room for improvement. For the reference, my
original ZEO-specific data corruption reproducer is here:~~

~~https://github.com/zopefoundation/ZEO/issues/155#issuecomment-577602842
https://lab.nexedi.com/kirr/wendelin.core/blob/ecd0e7f0/zloadrace5.py~~

**EDIT**: this fix now has corresponding test that should be coming in via https://github.com/zopefoundation/ZEO/pull/170 and https://github.com/zopefoundation/ZODB/pull/345.

/cc @d-maurer, @jamadden, @dataflake, @jimfulton

### DIFF
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -252,10 +252,19 @@ class Protocol(base.Protocol):
         message_id = (oid, tid)
         future = self.futures.get(message_id)
         if future is None:
-            future = asyncio.Future(loop=self.loop)
+            future = Fut()
             self.futures[message_id] = future
             self._write(
                 self.encode(message_id, False, 'loadBefore', (oid, tid)))
+            @future.add_done_callback
+            def _(future):
+                try:
+                    data = future.result()
+                except Exception:
+                    return
+                if data:
+                    data, start, end = data
+                    self.client.cache.store(oid, start, end, data)
         return future
 
     # Methods called by the server.
@@ -608,9 +617,6 @@ class Client(object):
                 future.set_exception(exc)
             else:
                 future.set_result(data)
-                if data:
-                    data, start, end = data
-                    self.cache.store(oid, start, end, data)
         elif wait_ready:
             self._when_ready(
                 self.load_before_threadsafe, future, wait_ready, oid, tid)
@@ -620,10 +626,7 @@ class Client(object):
     @future_generator
     def _prefetch(self, oid, tid):
         try:
-            data = yield self.protocol.load_before(oid, tid)
-            if data:
-                data, start, end = data
-                self.cache.store(oid, start, end, data)
+            yield self.protocol.load_before(oid, tid)
         except Exception:
             logger.exception("prefetch %r %r" % (oid, tid))
 
@@ -888,20 +891,25 @@ class ClientThread(ClientRunner):
                 raise self.exception
 
 class Fut(object):
-    """Lightweight future that calls it's callback immediately rather than soon
+    """Lightweight future that calls it's callbacks immediately rather than soon
     """
 
+    def __init__(self):
+        self.cbv = []
+
     def add_done_callback(self, cb):
-        self.cb = cb
+        self.cbv.append(cb)
 
     exc = None
     def set_exception(self, exc):
         self.exc = exc
-        self.cb(self)
+        for cb in self.cbv:
+            cb(self)
 
     def set_result(self, result):
         self._result = result
-        self.cb(self)
+        for cb in self.cbv:
+            cb(self)
 
     def result(self):
         if self.exc:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changelog
 5.2.3 (unreleased)
 ------------------
 
+- Fix data corruption due to race between load and external invalidations.
+  See `issue 155 <https://github.com/zopefoundation/ZEO/issues/155>`_.
 
 5.2.2 (2020-08-11)
 ------------------


navytux:
CI:

py36 [failed](https://travis-ci.com/github/zopefoundation/ZEO/jobs/498522599) with:

```
    Exception ignored in <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=False debug=False>>
    Traceback (most recent call last)
      File "/opt/python/3.6.7/lib/python3.6/asyncio/base_events.py", line , in __del__
        self.close()
      File "/opt/python/3.6.7/lib/python3.6/asyncio/unix_events.py", line , in close
        super().close()
      File "/opt/python/3.6.7/lib/python3.6/asyncio/selector_events.py", line , in close
        self._close_self_pipe()
      File "/opt/python/3.6.7/lib/python3.6/asyncio/selector_events.py", line , in _close_self_pipe
        self._remove_reader(self._ssock.fileno())
      File "/opt/python/3.6.7/lib/python3.6/asyncio/selector_events.py", line , in _remove_reader
        key = self._selector.get_key(fd)
      File "/opt/python/3.6.7/lib/python3.6/selectors.py", line , in get_key
        return mapping[fileobj]
      File "/opt/python/3.6.7/lib/python3.6/selectors.py", line , in __getitem__
        fd = self._selector._fileobj_lookup(fileobj)
      File "/opt/python/3.6.7/lib/python3.6/selectors.py", line , in _fileobj_lookup
        return _fileobj_to_fd(fileobj)
      File "/opt/python/3.6.7/lib/python3.6/selectors.py", line , in _fileobj_to_fd
        raise ValueError("Invalid file descriptor {}".format(fd))
    ValueError Invalid file descriptor -1
    <BLANKLINE>
```

uvloop [failed](https://travis-ci.com/github/zopefoundation/ZEO/jobs/498522609) with

```
RuntimeError: uvloop requires Python 3.7 or greater
An error occurred when trying to install /tmp/tmph6b7orkdget_dist/uvloop-0.15.2.tar.gz. Look above this message for any errors that were output by easy_install.
While:
  Installing test.
  Getting distribution for 'uvloop>=0.5.1'.
```

I believe both failures are unrelated to my patch.

The other 12 jobs [passed ok](https://github.com/zopefoundation/ZEO/pull/169/checks?check_run_id=2345573383).

src/ZEO/asyncio/client.py
```diff
@@ -888,20 +891,27 @@ def close(self):
                 raise self.exception
 
 class Fut(object):
-    """Lightweight future that calls it's callback immediately rather than soon
+    """Lightweight future that calls it's callbacks immediately rather than soon
     """
 
+    cbv = None
```
jmuchemb, line 897: Isn't `Fut()` usually followed by a call to `add_done_callback` ?

Then I'd replace with:
```py
def __init__(self):
    self.cbv = []
```
remove the `if self.cbv is None:` clause in `add_done_callback`, and the `if self.cbv:` lines in `set_*`.

src/ZEO/asyncio/client.py
```diff
@@ -620,10 +626,7 @@ def load_before_threadsafe(self, future, wait_ready, oid, tid):
     @future_generator
     def _prefetch(self, oid, tid):
         try:
-            data = yield self.protocol.load_before(oid, tid)
-            if data:
-                data, start, end = data
-                self.cache.store(oid, start, end, data)
+            yield self.protocol.load_before(oid, tid)
```
jmuchemb, line 629: Oh I missed that, thanks.

Is is tested ? I'm not asking about the race, but maybe not doing this change would break prefetch. Or would it just become dead code (i.e. data always false) ?

src/ZEO/asyncio/client.py
navytux: Good idea. Thanks for the advice.

src/ZEO/asyncio/client.py
navytux: You are welcome, and yes - prefetech is tested:

https://github.com/zopefoundation/ZEO/blob/3d90ed421b22b403ae32cf5d76ced53409479ecb/src/ZEO/tests/testZEO.py#L1071-L1113

What is removed is just a duplicate of cache update that is now living in `protocol.load_before`. It seems to be it would not be a bug to leave cache.store here because it would only store into the cache what was just stored into the cache by protocol.load_before itself. The wakup here seems to be immediate after switching load_before from `asyncio.Future` to `Fu`t, so there is seemingly no race window with invalidations. At least I could not make a data corruption with reintroducing this block back and injecting prefetches into my zloadrace.

Anyway, since prefech is just semantically "call loadBefore, but do not wait for it to complete", and `load_before_threadsafe` - that is serving `loadBefore` now is not doing cache update and delegates cache update to `protocol.load_before`, this change should be correct.

navytux:
Patch updated with `Fut` changes [suggested](https://github.com/zopefoundation/ZEO/pull/169#pullrequestreview-636125358) by @jmuchemb.

navytux:
( CI [failure](https://github.com/zopefoundation/ZEO/pull/169/checks?check_run_id=2354899951) is just with uvloop _"RuntimeError: uvloop requires Python 3.7 or greater"_ and unrelated to my patch )

navytux:
This fix now comes with corresponding test: see https://github.com/zopefoundation/ZODB/pull/345#issuecomment-821316526 and https://github.com/zopefoundation/ZEO/pull/170.

d-maurer (approved):
The proposed fix lets the Travis tests for #167 (which formerly reliably revealed a race condition problem) succeed. Thus, it apparently resolves at least that problem.

I suggest to add a change log entry to document the fix.

d-maurer:
See "https://github.com/zopefoundation/ZEO/pull/167/commits/24f4bb785cbf06613a8b1a2923aa8e5b6e906898" to resolve the `uvloop` Travis problem.

navytux:
Thanks, @d-maurer. I've added the changelog entry as you suggests. Note, that this fix now has correposponding test that should be coming in via https://github.com/zopefoundation/ZEO/pull/170 and https://github.com/zopefoundation/ZODB/pull/345.

If ok, let's merge all this bits.

navytux:
(and good to have that uvloop problem now fixed)

navytux:
CI is ok modulo uvloop. But uvloop fix is coming in separately. (@d-maurer, btw, maybe it makes sense to cherry-pick your "fix uvloop" patch to master without waiting for https://github.com/zopefoundation/ZEO/pull/167 to be merged first.)

d-maurer:
Kirill Smelkov wrote at 2021-4-20 04:26 -0700:
>CI is ok modulo uvloop. But uvloop fix is coming in separately. ***@***.***, btw, maybe it makes sense to cherry-pick your "fix uvloop" patch to master without waiting for https://github.com/zopefoundation/ZEO/pull/167 to be merged first.)

I would not mind.

navytux:
> I would not mind.

Thanks, @d-maurer. I've applied your uvloop-fix patch to master: https://github.com/zopefoundation/ZEO/commit/707316c4.

navytux:
- patches rebased onto master to get uvloop fix;
- patches squashed into one to update commit message with properly referring to where corresponding test is coming.

navytux:
I believe it's time to merge this fix.

navytux:
This patch fixes data corruption. It was merged to master almost 4 months ago. Can we please make a new ZEO release to get this fix available out of the box?

Thanks beforehand,
Kirill

/cc @jamadden, @icemac, @dataflake, @jugmac00

dataflake:
I'll create a release

navytux:
Thanks, @dataflake.

dataflake:
ZEO 5.2.4 is now published

dataflake:
oops, 5.2.3 I meant

navytux:
@dataflake, thanks again. I appreciate your help.

PR #170, opened 2021-04-16
tests: Factor "Open another ClientStorage to the same server" into generic _new_storage_client()

navytux:
This allows ZODB tests to recognize ZEO as client-server storage and so
make "load vs external invalidate" test added in
https://github.com/zopefoundation/ZODB/pull/345 to reproduce data
corruption problem reported at
https://github.com/zopefoundation/ZEO/issues/155.

For the reference: that problem should be fixed by
https://github.com/zopefoundation/ZEO/pull/169.

We drop

    # It's hard to find the actual address.
    # The rpc mgr addr attribute is a list.  Each element in the
    # list is a socket domain (AF_INET, AF_UNIX, etc.) and an
    # address.

note because at the time it was added (81f586c4) it came with

    addr = self._storage._rpc_mgr.addr[0][1]

but nowdays after 03867188 getting to server address is just by ClientStorage._addr.

### DIFF
--- a/src/ZEO/tests/testZEO.py
+++ b/src/ZEO/tests/testZEO.py
@@ -190,9 +190,7 @@ class MiscZEOTests(object):
         self._dostore(data=obj)
 
     def checkZEOInvalidation(self):
-        addr = self._storage._addr
-        storage2 = self._wrap_client(
-            ClientStorage(addr, wait=1, **self._client_options()))
+        storage2 = self._new_storage_client()
         try:
             oid = self._storage.new_oid()
             ob = MinPO('first')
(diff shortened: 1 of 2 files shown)


navytux:
CI: besides usual "uvloop requires Python 3.7" there is `zeo-fan-out.test` [failure](https://travis-ci.com/github/zopefoundation/ZEO/jobs/499139955) on pypy3, that I belive is unrelated to my patch. Other [12 jobs pass ok](https://github.com/zopefoundation/ZEO/pull/170/checks?check_run_id=2364305401).

src/ZEO/tests/testZEO.py
```diff
@@ -262,6 +259,15 @@ def setUp(self):
...
         self._storage.registerDB(DummyDB())
 
+    # _new_storage_client opens another ClientStorage to the same storage server
+    # self._storage is connected to. It is used by both ZEO and ZODB tests.
+    def _new_storage_client(self):
+        client = ZEO.ClientStorage.ClientStorage(
+            self._storage._addr, wait=1, **self._client_options())
+        client = self._wrap_client(client)
+        client.registerDB(DummyDB())
+        return client
```
d-maurer, line 269: Should the storage really already register with a DB? In some situation, we might want an `MVCCAdapter` to be registered.

navytux: @d-maurer, thanks for feedback. I've thought about that too, and my logic here is:

- the original code being factored-out had that `registerDB(DummyDB())`. There is also the same `registerDB(DummyDB())` in `GenericTestBase.setUp` which prepares `._storage` client for practically most of the tests:

https://github.com/zopefoundation/ZEO/blob/707316c49984ffbf8f2347611de5abdc3230dd3f/src/ZEO/tests/testZEO.py#L236-L263

- the `DB.registerDB`, when called for the second time, forgets what was registered first. This way when/if MVCCAdapter registers into ClientStorage, the registration of DummyDB is dropped.

The code in `_new_storage_client` is this OK, and is in line with the way `ClientStorage` for main `._storage` is prepared.

d-maurer (approved):

navytux:
Thanks, @d-maurer

PR #173, opened 2021-04-23
improve log message when client cache is out of sync with server

### LINKED ISSUE #142: Improve log and exception message
When there is a newer transaction id on the client than on the server - for instance after a backup has been restored - following message gets logged:
"ClientStorageError: Client has seen newer transactions than server!"

The admin is left with this message and probably does not know how to recover a working state - as Zope is no longer usable.

What worked for me - and maybe is not always the best solution was to remove the zec file, cf 
https://docs.plone.org/3/en/manage/troubleshooting/transactions.html#zeostorage-client-has-seen-newer-transactions-than-server

There may be other reasons for this error and maybe other recovery steps make more sense then.

From a quick glance tox is not setup to run coverage (see #123 ), and from looking at the code and the tests it seems the branch ("Cache newer than server") is untested.

**Update**
The branch gets covered by 
https://github.com/zopefoundation/ZEO/blob/a6c1ffcefda76354b07e56af463d5da8e90cf60c/src/ZEO/asyncio/tests.py#L437
Was hard to spot, as the test does neither check for the AssertionError nor for the log message.

https://github.com/zopefoundation/ZEO/blob/bd7cd066a49d7090263211574aa0a0a38010cd65/src/ZEO/asyncio/client.py#L488-L493

Also, the tests for ZEO seems to be quite flaky, cf #140 

@jamadden made some good suggestions how to proceed:
https://github.com/zopefoundation/ZEO/pull/139#discussion_r278190611

[Removed wrong assumption about assert / AssertionError]

jugmac00:
This fixes #142

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@ Changelog
 - Fix data corruption due to race between load and external invalidations.
   See `issue 155 <https://github.com/zopefoundation/ZEO/issues/155>`_.
 
+- Improve log message when client cache is out of sync with server.
+  See `issue 142 <https://github.com/zopefoundation/ZEO/issues/142>`_.
+
 5.2.2 (2020-08-11)
 ------------------
 
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -497,7 +497,10 @@ class Client(object):
                 elif cache_tid > server_tid:
                     self.verify_result = "Cache newer than server"
                     logger.critical(
-                        'Client has seen newer transactions than server!')
+                        'Client cache is out of sync with the server. '
+                        'Verify that this is expected and then remove '
+                        'the cache file (usually a .zec file) '
+                        'before restarting the server.')
                     raise AssertionError("Server behind client, %r < %r, %s",
                                          server_tid, cache_tid, protocol)
                 elif cache_tid == server_tid:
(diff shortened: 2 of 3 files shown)


jugmac00:
The new log message is not asserted by a test, as I am not aware how to to this with `zope.testrunner`.

If anybody could give me a hint, I will gladly update the current test ( `test_bad_server_tid`) to also check for the message.

dataflake (approved):

dataflake:
Is a test really needed? You know the error message shows up from your own experience. Just changing the text won't change the fact that it shows up. Testing the text of the message itself seems pointless. The only thing I would test, even if it's just manually in `pdb`, is the string representation of `self.cache`. Is that always a path?

jugmac00:
> Is a test really needed? You know the error message shows up from your own experience. Just changing the text won't change the fact that it shows up. Testing the text of the message itself seems pointless.

While the local tests passed as tox does not run all tests, at CI I got another error and looks like Jim used some kind of doctest to test the logging message!

https://github.com/zopefoundation/ZEO/blob/43a82469fb9ee4717af9f922ba0d2759e5201e13/src/ZEO/tests/testZEO.py#L1150

I'll fix the test in a minute.

> The only thing I would test, even if it's just manually in `pdb`, is the string representation of `self.cache`. Is that always a path?

Jason gave me the tipp with `self.cache`, so I guess its valid. During the test you do not get a file, as it's all in memory.

I try to replay the doctest in a shell with a real `Data.fs` and have a look at the value.

jugmac00:
Thanks, Jens. I removed the reference to `self.cache` as I was not able to trigger the message with a file based cache in a reasonable amount of time, and for a mem-based cache it would have been misleading.

jugmac00:
the test failure is unrelated and already reported as #179

PR #174, opened 2021-04-23
let `lastTransaction` change only after invalidation processing

### LINKED ISSUE #166: Another data corruption due to concurrency bug  (ZEO5 regression; no external invalidations)

d-maurer:
... as required by `ZODB >= 5.6`.
Fixes #166.

This PR is a cleanup of and replacement for #167:
 * The required effect is achieved via locking - to satisfy the requirement of both `ZODB >= 5.6` and that of the  `check_tid_ordering_w_commit` tests.
 * Source code comments explain the locking and why I think that locking protecting cache invalidations is not necessary.
 * The race condition tests have been removed because upcoming ZODB versions will contain comparable (generic) tests.


navytux:
@d-maurer, thanks.

So focusing on that `cache.setLastTid()` and `f()` must be done together atomically, this PR looks good to me.

The logic that cache updates are ok to be done out of atomic section, because cache is updated only from asyncio thread, and other threads use only cache.loadBefore with tid <= current cache.lastTid is also sound, though somewhat uncomfortable (see https://github.com/zopefoundation/ZEO/pull/167#issuecomment-821321036 once again).

For the reference: I've found brakeage of this invariant, because ZEO ClientStorage (a normal thread) wants to invalidate the cache on conflicting vote:

https://github.com/zopefoundation/ZEO/blob/3111c7e055348e450c9bbf9a7302e8668c3dfa58/src/ZEO/ClientStorage.py#L774-L785

This was added in https://github.com/zopefoundation/ZEO/commit/062cbecc and it generally looks wrong to do such invalidation from outside of asyncio thread because it looks like it can potentially hit order rearranging scenarious with e.g. from-server invalidations and/or loads from other client threads.

The client cache was made to be thread-safe right after that commit in https://github.com/zopefoundation/ZEO/commit/5e6fd05f. Both those changes were part of implementing client-side conflict-resolution in https://github.com/zopefoundation/ZEO/pull/38.

Atomically processing invalidations won't help here, but I wanted to share that finding while we are on this topic.


> There is still one point I have not yet fully understood.
In the discussion about the "loadBefore" race I got the impression
that part of message processing could be out of order.
I think this part refered to the processing of "done callbacks", not
the initial message processing. But, I would like to understand
this better before the merge.


Before https://github.com/zopefoundation/ZEO/commit/e277ba09 (the fix for https://github.com/zopefoundation/ZEO/issues/155) cache updates were only scheduled to be eventually performed after receiving loadBefore reply. After that patch they are done immediately after receiving loadBefore reply. In other words now all loadBefore messages should be processed completely in-order. And it is not only about loadBefore, because before e277ba09 asyncio.Future was used only in that function and after that patch now it is `Fut` that is used everywhere. And the purpose of `Fut` is to invoke "done" callbacks immediately:

https://github.com/zopefoundation/ZEO/blob/3111c7e055348e450c9bbf9a7302e8668c3dfa58/src/ZEO/asyncio/client.py#L893-L895


In summary: I approve this PR. Let's move on and if we find more problems with e.g. non-atomicity of invalidations processing, we can address them in follow-up steps.

Thanks again for you work to bring ZEO5 client into better shape,

Kirill

navytux (approved):

d-maurer:
The  `drop_cache_rather_than_verify` failure in "https://github.com/zopefoundation/ZEO/pull/174/checks?check_run_id=2435058350" is caused by one of the known race conditions -- and not related to this PR.

d-maurer:
I had to cancel the `pypy3` job -- it did not finish after more than 1 hour (usual runtime 5 to 10 minutes). I suspect a race condition leading to a deadlock -- unfortunately, there is no detailed information (e.g. tracebacks, hanging test, etc) for the analysis.

navytux:
@d-maurer, thanks again. Probably this particular deadlock was not due to this, but since we are talking about deadlock scenarious, maybe https://github.com/zopefoundation/ZEO/pull/182#issuecomment-826630195 is related:

> In other words by not letting that code to run, we, in the best case, leak resources, and in the worst create potential deadlock scenarios (e.g. some caller is waiting for a future to complete one way or another after closing the link/protocol, and the future will stay in not done state indefinitely).

Just FYI.

jmuchemb:
I am annoyed with the merge of this PR. I can understand to quickly master with something safe but I'm afraid we stop here.

First of all, I think the `with cache._lock` is useless, in which case the check_tid_ordering_w_commit test should be changed. I didn't take the time to read all comments carefully, so maybe some comment proves that I'm wrong.

Anyway, it seems we all agree that a lock outside the cache would be better than an internal one. So my question is: is there any plan to change this ?

d-maurer:
Julien Muchembled wrote at 2021-5-4 23:41 -0700:
> ...
>First of all, I think the `with cache._lock` is useless

It is not useless as it allowed to fulfill
the `ZODB >= 5.6` requirement that `lastTransaction` changes
only after invalidation processing
and at the same time let the `check_tid_ordering_w_commit`
succeed (as you had first noticed).
It does this by delaying the `lastTransaction` return
until the callback has run.
Apparently, other storages, too, use locking to achieve this.

> ...
>Anyway, it seems we all agree that a lock outside the cache would be better than an internal one.

I see this only as cosmetics: I would not introduce a new lock
just to protect `lastTransaction` and the `setLastTid`
in `tpc_finish_threadsafe`.

>So my question is: is there any plan to change this ?

Not from me.

jmuchemb:
> Apparently, other storages, too, use locking to achieve this.

For NEO, not really. If I wanted to micro-optimize NEO/ZODB5 as much as possible, lastTransaction would not be protected by the lock and I'd have to change `check_tid_ordering_w_commit`.

When I write that the `with cache._lock` would be useless, it's for the proper functioning of ZEO. `check_tid_ordering_w_commit` is stricter than the ZODB interface and it's a bad idea to add `with cache._lock` *only* to make the test (as currently written) pass.

> > Anyway, it seems we all agree that a lock outside the cache would be better than an internal one.
>
> I see this only as cosmetics: I would not introduce a new lock just to protect `lastTransaction` and the `setLastTid` in `tpc_finish_threadsafe`.

It's not just cosmetics and I'm not talking about a new lock: see my reply to the last quote in https://github.com/zopefoundation/ZEO/pull/167#issuecomment-819917751

At the end, you'd have a single `with lock:` that spans several lines of `tpc_finish_threadsafe`, no other lock taken during subcalls.  Whether `setLastTid` is called while the lock is still taken would not make much difference anymore and we could keep `check_tid_ordering_w_commit` as it is.

d-maurer:
Julien Muchembled wrote at 2021-5-5 00:59 -0700:
> ...
>It's not just cosmetics and I'm not talking about a new lock: see my reply to the last quote in https://github.com/zopefoundation/ZEO/pull/167#issuecomment-819917751
>
>At the end, you'd have a single `with lock:` that spans several lines of `tpc_finish_threadsafe`, no other lock taken during subcalls.  Whether `setLastTid` is called while the lock is still taken would not make much difference anymore and we could keep `check_tid_ordering_w_commit` as it is.

I am satisfied with the current solution (race condition
involving `tpc_finish` no longer observed; all tests including
race condition tests pass). If you think to have a better one implement it.

Issue #123, opened 2018-12-14
Add coverage tests


icemac:
It would be nice to test the coverage of this project and to kept track of it using https://coveralls.io.

jamadden:
+1 

The tests are run in parallel subprocess, and that does make a substantial difference so I doubt we can drop that. So correctly measuring coverage requires special setup [like ZODB used to use](https://github.com/zopefoundation/ZODB/commit/d797b193feece3069a7f461dfd963469009ca968)

PR #175, opened 2021-04-23
Switch from Travis CI to GitHub Actions for testing

dataflake:
Travis CI can be considered broken at this point. After they switched their business model and dropped the free unlimited use for Open Source communities they made it extremely hard to get enough free runs.

This PR moves all those test configurations into GitHub Actions: https://github.com/zopefoundation/ZEO/actions/workflows/tests.yml


jugmac00 (approved):
LGTM! I had to restart CI twice to make all tests pass. We have some flaky tests.

d-maurer:
Jürgen Gmach wrote at 2021-4-23 13:45 -0700:
***@***.*** approved this pull request.
>
>LGTM! I had to restart CI twice to make all tests pass. We have some flaky tests.

There is a known race condition which should be fixed by #174.

d-maurer:
> Jürgen Gmach wrote at 2021-4-23 13:45 -0700: ***@***.*** approved this pull request.
> LGTM! I had to restart CI twice to make all tests pass. We have some flaky tests.
> There is a known race condition which should be fixed by #174.

Likely, I was wrong: the race condition test related to #174 comes with an upcoming (not yet released) ZODB version which almost surely is not used in the current setup.The failing tests you have observed likely indicate new race conditions.

navytux:
@dataflake, first of all thanks for caring about ZEO CI.

Do you think it could be possible to also add an entry into the matrix to verify ZEO against `ZODB@master`?

Doing so, in addition to verifying ZEO against ZODB release, will activate ZEO to run tests that reproduce data corruptions from  https://github.com/zopefoundation/ZEO/issues/155 and https://github.com/zopefoundation/ZEO/issues/166 (see https://github.com/zopefoundation/ZEO/pull/167#issuecomment-824602061 for details), and it would also keep ZEO being tested for in-progress ZODB development.

Should we have such `ZEO-ZODB@master` coverage before, it is very likely that data corruption bug associated with https://github.com/zopefoundation/ZEO/issues/166 could be prevented from being introduced at all. The reason here is that bug started to appear only with ZODB 5.6, because during development of https://github.com/zopefoundation/ZODB/pull/307  something was overlooked (details: https://github.com/zopefoundation/ZEO/issues/166#issuecomment-815625731). Having ZEO-ZODB@master coverage would likely help to indicate that overlook with test failures, and would make developers to look into why things fail _before_ ZODB 5.6 was released.

Thanks beforehand,
Kirill

/cc @d-maurer, @jamadden, @jmuchemb

d-maurer (approved):
Thank you for your quick action!

I do not yet know git actions at all and am thus likely not optimal to review this PR. But the known ZEO test jobs are run and able to detect problems. That means a big improvement.

dataflake:
@navytux I'll add that ZODB master test separately and will merge this now so it's available. Travis CI has failed us again and is not running at all now.

navytux:
> @navytux I'll add that ZODB master test separately

Thanks, @dataflake.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,61 @@
...
+          ${{ runner.os }}-pip-${{ matrix.config[0] }}-
+          ${{ runner.os }}-pip-
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
+    - name: Test
+      run: tox -e ${{ matrix.config[1] }}
+    - name: Coverage
+      if: matrix.config[1] == 'coverage'
```
jugmac00, line 56: I have not used gha with matrix, but I think this gets never true.

Sorry, just noticed this now as I saw we have an open issue about adding coverage, see #123

dataflake: It is there for the moment when we have coverage testing.

Issue #176, opened 2021-04-24
`trollius` bug causes problems for ZEO testsuite in Python 2 environments


d-maurer:
The `py27-msgpack1` job of "https://github.com/zopefoundation/ZEO/pull/175/checks?check_run_id=2425420460" has `ZODB.tests.testblob.ClientStorageNonSharedBlobsBlobUndoTests.testUndoAfterConsumption` failed:
```
Error in test testUndoAfterConsumption (ZODB.tests.testblob.ClientStorageNonSharedBlobsBlobUndoTests)
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/ZODB/tests/testblob.py", line 236, in testUndoAfterConsumption
    database.close()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/ZODB/DB.py", line 645, in close
    self._mvcc_storage.close()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/ZODB/mvccadapter.py", line 68, in close
    self._storage.close()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/ZEO/tests/testZEO.py", line 1739, in close
    ClientStorage.close(self)
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/ZEO/ClientStorage.py", line 305, in close
    self._server.close()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/ZEO/asyncio/client.py", line 888, in close
    self.loop.call_soon_threadsafe(self.loop.stop)
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/trollius/base_events.py", line 499, in call_soon_threadsafe
    self._write_to_self()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-msgpack1/lib/python2.7/site-packages/trollius/selector_events.py", line 166, in _write_to_self
    wrap_error(csock.send, b'\0')
[...]

PR #180, opened 2021-04-24
Add automated tests against the ZODB master branch

### LINKED ISSUE #177: Add a test against ZODB master branch to GH actions

dataflake:
Fixes #177 

This PR does the following:

- cut down `tox.ini` environment descriptions by using shorthand notation
- ignore `ResourceWarnings` on `tox` runs for Python 3
- add an automated  GitHub Actions test run using Python 3.7 against the ZODB ``master`` branch.

I didn't know how to express something like "the dependency ZODB needs to be checked out" in `tox.ini` itself, so I took a detour via a dedicated `zc.buildout` configuration. Let me know if you have a more elegant solution.


navytux:
@dataflake, thanks for doing this.

> I didn't know how to express something like "the dependency ZODB needs to be checked out" in tox.ini itself, so I took a detour via a dedicated zc.buildout configuration. Let me know if you have a more elegant solution.

Here is example how one can add environment to tox test matrix to verify things wrt ZODB checkout:

https://lab.nexedi.com/nexedi/zodbtools/commit/c50bfb00

Another thing, noted during working on recent fixes to ZEO, is that concurrency bugs are much more likely to trigger on python2. This way I would suggest to have both `py27-zodb-master` and `py37-zodb-master`.

d-maurer:
Jens Vagelpohl wrote at 2021-4-24 05:26 -0700:
> ...
>I didn't know how to express something like "the dependency ZODB needs to be checked out" in `tox.ini` itself, so I took a detour via a dedicated `zc.buildout` configuration. Let me know if you have a more elegant solution.

Usually, "pip" is used to install requirements.
"pip" supports "VCS project urls" as package specifiers
and "git" is one of the supported "Version Control Systems".

"VCS project urls" have the form: *scheme*+*url*
*url* is interpreted by the the VCS identified by *scheme*.
Thus, it can use any means supported by the VCS to identify
branches or revisions.

`git+https://github.com/pypa/pip.git` is an example for a git project url.
Apparently (from "https://matiascodesal.com/blog/how-use-git-repository-pip-dependency/"), "git" is using @ suffixes to specify a revision.
Thus ***@***.***` should specify
the `master` branch.

jugmac00:
Here are some more examples:
https://pip.pypa.io/en/stable/cli/pip_install/#vcs-support

dataflake:
Thank you all for your help, I managed to drop the buildout step and make the PR much smaller. Please take a look and leave a review when you have a minute.

navytux:
Thanks, @dataflake. I have not verified in detail, but the changes look reasonable to me. The failure in `check_race_loadopen_vs_local_invalidate`  in py27-zodbmaster is expected as we don't have https://github.com/zopefoundation/ZEO/pull/174 merged yet. The failure with "bad file descriptor" is also known (see https://github.com/zopefoundation/ZEO/issues/176). In general I believe it is ok to merge this.

navytux (approved):

Issue #165, opened 2021-02-12
Support for Python 3.8 (broken?)


dataflake:
Right now the ZEO package declares support for Python up to 3.7. Trying to run the tests on Python 3.8 produces two errors, see below. Basically, the `asyncio.Future` object `future` raises the exception because it is in status `cancelled`. The asynchronous operation "hides" the actual reason and I cannot figure it out.

Other than that there are some other failures on 3.8 that appear to be race and/or timing issues.

I cannot judge right now if running a ZEO server on Python 3.8 is "safe" in the context of e.g. a Zope 5 install under Python 3.8 or 3.9.

```
$ bin/test -tmultiple_addresses
Running zope.testrunner.layer.UnitTests tests:
  Set up zope.testrunner.layer.UnitTests in 0.000 seconds.


Error in test test_multiple_addresses (ZEO.asyncio.tests.ClientTests)
Traceback (most recent call last):
  File "/home/jens/src/Python-3.8.6/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/home/jens/src/Python-3.8.6/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/home/jens/src/Python-3.8.6/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/jens/src/ZEO/src/ZEO/asyncio/tests.py", line 430, in test_multiple_addresses
    self.respond(1, None)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/tests.py", line 119, in respond
    self.loop.protocol.data_received(
  File "/home/jens/src/ZEO/src/ZEO/asyncio/base.py", line 128, in data_received
    self.message_received(collected)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 232, in message_received
    future.set_result(args)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 904, in set_result
    self.cb(self)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 45, in _
    f = gen.send(result)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 202, in finish_connect
    self.client.registered(self, server_tid)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 447, in registered
    self._clear_protocols(protocol)
[...]

Issue #179, opened 2021-04-24
flaky test:  `drop_cache_rather_than_verify`


jugmac00:
see https://github.com/zopefoundation/ZEO/pull/178/checks?check_run_id=2425920984

This error does not happen all the time = flaky test.

```
Failure in test /home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/tests/drop_cache_rather_than_verify.txt
Failed doctest test for drop_cache_rather_than_verify.txt
  File "/home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/tests/drop_cache_rather_than_verify.txt", line 0

----------------------------------------------------------------------
File "/home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/tests/drop_cache_rather_than_verify.txt", line 146, in drop_cache_rather_than_verify.txt
Failed example:
    for e in events:
        print(e)
Expected:
    (2, '', 'StaleCache')
Got nothing

...../opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/subprocess.py:689: ResourceWarning: unclosed file <_io.BufferedReader name=4>
  _cleanup()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/nagios.py:140: ResourceWarning: unclosed <socket.socket fd=9, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('0.0.0.0', 51580)>
  addr, options.output_metrics, options.status_path, options.time_units)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
......../home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/asyncio/client.py:880: ResourceWarning: unclosed resource <TCPTransport closed=False reading=True 0x7f1560003d90>
  loop.close()
/home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/_forker.py:181: ResourceWarning: unclosed file <_io.BufferedRandom name='./1-1.zec'>
  gc.collect()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
[...]

PR #182, opened 2021-04-24
#165 more accurate asyncio.CancelledError handling

vernans:
#165 is the related issue for this.


dataflake (approved):

dataflake:
@navytux Do you have any input on this change for Python 3.8+ support?

navytux:
@dataflake, thanks for asking. First of all I don't think I should be in good position to review asyncio code because it is, imho unneccessarily, very twisty and it is easy for a human to miss potential code flows(*). Still I've tried to have a look at the changes and at @vernans explanation in https://github.com/zopefoundation/ZEO/issues/165#issuecomment-826166479 (btw, @vernans, thanks for the explanation and for trying to fix things).

To me the intention to catch "canceled" error when closing several in-flight-being-established protocols is correct. Indeed we spawned several protocols to be established simultaneously, and when the first one "wins", we cancel others

https://github.com/zopefoundation/ZEO/blob/1070adcbd17b83b9515c43b48ca38687783793e4/src/ZEO/asyncio/client.py#L450-L456

However the way to handle "canceled" error in `_clear_protocols` by wrapping calls to `Protocol.close()` into `try/except asyncio.CancelledError` seems to be not completely correct to me. As https://github.com/zopefoundation/ZEO/issues/165#issue-807061922 shows, because our `Fut` invokes callbacks immediately, the hit into raised `CancelledError` is happenning right from `self._connection.cancel()` call inside `Protocol`:

https://github.com/zopefoundation/ZEO/blob/1070adcbd17b83b9515c43b48ca38687783793e4/src/ZEO/asyncio/client.py#L95-L98

This means that if we wrap with try/except CancelledError the call to `Protocol.close` itself, the rest of the `Protocol.close` after `._connection.cancel()` call won't be executed. And the rest of the code closes transport and "cancels" already in-flight futures over the protocol that is being closed:

https://github.com/zopefoundation/ZEO/blob/1070adcbd17b83b9515c43b48ca38687783793e4/src/ZEO/asyncio/client.py#L98-L102

In other words by not letting that code to run, we, in the best case, leak resources, and in the worst create potential deadlock scenarios (e.g. some caller is waiting for a future to complete one way or another after closing the link/protocol, and the future will stay in not done state indefinitely).

So I would suggest to rework the CancelledError handling by moving it closer to origin of the cancellation - to around `._connection.cancel()` call itself - and wrap that call with the try/except CancelledError. This way the code to close transport and finish in-flight futures would be also run. The code would also seem to be more clear to me after this change: we cancel things, and we expect at exactly that place that CancelledError would happen. The cause and corresponding error handling would be placed in the same single place making it easier to track the logic.

Going that route I would suggest to also do the following: if I understand correctly a `CancelledError` can happen only when things are explicitly cancelled. This way other places that trigger cancellation should be also reviewed. Besides `Protocol.close` discussed above, there is only one other place where explicit cancellation is issued - in `Protocol.connection_lost`:

https://github.com/zopefoundation/ZEO/blob/1070adcbd17b83b9515c43b48ca38687783793e4/src/ZEO/asyncio/client.py#L143-L148

So it means the user of `Protocol.heartbeat_handle` and the invokers of `Protocol.futures` have to be reviewed for handling CancelledError correctly as well. The latter means practically all RPC calls from client to server...

For `protocol.fut('lastTransaction')` vs `ClientDisconnected` in verify - I would suggest to do the following instead: just move up the try/except that is _already_ covering the code in verify a bit up - to also cover that lastTransaction call. This try / except already catches `except Exception` and does `self.register_failed(protocol, exc)` on a failure:

https://github.com/zopefoundation/ZEO/blob/1070adcbd17b83b9515c43b48ca38687783793e4/src/ZEO/asyncio/client.py#L488-L535

So if we just move its beginning `try` a bit up in here to also cover `lastTransaction` call here:

https://github.com/zopefoundation/ZEO/blob/1070adcbd17b83b9515c43b48ca38687783793e4/src/ZEO/asyncio/client.py#L485-L488

it will work out automatically. It will work correctly because `ClientDisconnected` is inheriting as

    ClientDisconnected -> ClientStorageError -> StorageError -> POSError -> Exception

It is also more logical to wrap the whole operation in verify in such a try/except block, so that if _anything_ at _any_ point fails, we mark the whole operation as failed.




Hope this helps a bit, and, once again I might be missing or misunderstanding something.

Kirill

(*) my preferrence is to use communication sequential processes instead (see for gevent/greenlet/pygolang/...)

vernans:
@navytux 
Thanks for the review. I tried to focus on restoring the old behavior in python3.8. Catching the error earlier would have been a breaking change. Nevertheless I also stumbled over the issue you described. (early exiting and not excecuting code because of CancelledError) It was the default behavior for years now, if the implications are this grave maybe we should open a new bug ticket about this. Maybe this needs further explanation and communication. _clear_protocols seemed to be the perfect spot, as it is an internal method.

I thought to have reviewed all places that call Future.exception() or Future.result() (the only methods to throw CancelledError). Those are mostly not catching exceptions or at least rethrowing them. Also most of them are class Fut(). Only for self.connected and self._connecting, concurrent.futures.Future is used. It would be possible to throw CancelledError by Fut, but only if a caught Exception of concurrent.futures.Future was put on Fut by calling Fut.set_exception, which wasn't the case as I looked over the code. I don't think the problem is as big as you describe it. 

For the ClientDisconnected Error caught in verify ­- sure! I mentioned it in the comment below, because I wasn't bold enough to put things into a try Exception clause myself. If you say so, we can do it.

navytux:
@vernans, thanks for feedback.

I cannot tell about ZEO4, but about ZEO5/ZODB5 what is clear is that it was really buggy for at least several years. Please look at least at data-corruption level ZODB4/ZEO4->ZODB5/ZEO5 regressions described in https://github.com/zopefoundation/ZEO/issues/155, https://github.com/zopefoundation/ZEO/issues/166 and https://github.com/zopefoundation/ZODB/issues/290. Even though those were recently fixed, CI still shows that deadlocks (see e.g. https://github.com/zopefoundation/ZEO/pull/174#issuecomment-826538354) and other concurrency bugs (see e.g. https://github.com/zopefoundation/ZEO/issues/179, https://github.com/zopefoundation/ZEO/issues/176) still remain.

From this point of view, what are the upsides of preserving previous buggy behaviour detouring from straight logic instead of fixing it while we are working on the code because a change in environment highlighted problems?

I'm asking that question to myself, and I cannot find any good answer...

About verify and lastTransaction. Does lastTransaction RPC could raise only ClientDisconnected error? Can it be an exception sent from the server as RPC reply? In other words - can it be _any_ error? To me the logic in verify was to mark verify as failed if there was any error, not only just disconnection, and it was just a thinko that first lastTransaction call was inserted outside of that try/except.

Kirill

d-maurer:
Kirill Smelkov wrote at 2021-4-26 11:39 -0700:
> ...
>About verify and lastTransaction. Does lastTransaction RPC could raise only ClientDisconnected error? Can it be an exception sent from the server as RPC reply? In other words - can it be _any_ error? To me the logic in verify was to mark verify as failed if there was any error, not only just disconnection, and it was just a thinko that first lastTransaction call was inserted outside of that try/except.

My recent code analysis of `ZEO.async.client` gave me the impression
that all RPC calls can return exceptions which occured on the server.
They are handled by calling `set_exception` on the future
-- i.e. in analogy to the handling of a normal result.
In my view, there is no need to use "try ... except" to handle those
exceptions.

`ClientDisconnected` is a different case: this error is detected
in the client and affects all outstanding RPC calls;
it needs to be properly propagated as those RPC calls are "broken"
and the callers must be informed.

dataflake:
The last failing test seems to fail not on functionality but on its text output due to an ignored exception deep in the bowels of Python's async code. I can't tell if that's important or not.

dataflake:
I am guessing there is another race condition. The `py39` test works fine in a local Linux sandbox (RHEL/CentOS/Rocky 8).

dataflake:
@d-maurer Just FYI, I just released ZODB 5.7.0, which will get pulled in for the tests from now on.

d-maurer:
I currently have the following theory about the race condition: for normal processing, `runzeo.ZEOServer` closes the (asyncio) loop as part of its `loop_forever`. However, if called with `--test`, `loop_forever` is essentially empty. As a consequence, the loop is not closed explicitly but implicitly during Python shutdown. It is known that things during shutdown do not happen deterministically: the race happens when the socket gets closed before the loop is deleted.

I have added an explicit event loop close for the `--test` case (there might be a better solution). All tests pass now.

icemac:
@navytux Do you think you can review this PR?

dataflake:
@icemac Well, @navytux has reviewed the PR, he just hasn't given a thumbs up yet. I have a feeling this is stuck in a situation where @navytux is suggesting improvements while @vernans wanted to stick as closely to the existing code to minimize risks. The improvement suggestions have ballooned a little bit to the point where I can't tell if the suggestions would fix real bugs or "just" improve code behavior. In addition, the discussion is happening in two places, here and in #165 , and I have a hard time tracking all that.

I think it would be great if @navytux could just answer the simple question if the PR as it stands is harmful/risky or not. If it doesn't add additional risk and just continues questionable code behavior it should be merged. Improving code behavior should be split out so this PR gets out of the discussion weeds.

navytux:
@icemac, @dataflake, thanks for feedback and for asking. I indeed had reviewed this PR and I believe what I wrote in https://github.com/zopefoundation/ZEO/pull/182#issuecomment-826630195 and in https://github.com/zopefoundation/ZEO/pull/182#issuecomment-827060623 still hold true and need to be addressed.

In my view addressing that is not just for the sake of it, but should fix real correctness problems (in the best case leakage of resources or deadlocks on an error codepath, if I remember correctly).

Given a large - almost 1 year - pause, after I posted my feedback, I believe there is no interest from author of this PR to address it (@vernans, please correct me if I'm wrong).

So I do not want to block progress and insist that this PR is merged only after the issues, I highlighted, are fixed. Maybe what should be done is to merge this PR as is and create an issue to address the problems I highlighted later. However personally I feel not very good about that.

Kirill

dataflake:
@navytux Would you be willing to implement changes that make you more comfortable right on this branch? If not I fear we're just plain stuck. I don't know if @vernans is interested anymore.

navytux:
@dataflake, thanks, that indeed might be the way to go. I've incorporated one easier part of my suggestions into hereby PR.

I was going to do similarly with what I suggested in first part of my message in https://github.com/zopefoundation/ZEO/pull/182#issuecomment-826630195 , but discovered, what I believe is, a more significant problem with both this PR and how ZEO/asyncio is tested at all:

It was said in https://github.com/zopefoundation/ZEO/issues/165#issuecomment-826166479 that we use Fut instead of asyncio.Future future. But this is the case for doing most of RPC calls, while e.g. Protocol._connecting is regular full asyncio.Future. For the record, the main difference in between Fut and asyncio.Future is that Fut "calls it's callbacks immediately rather than soon"

https://github.com/zopefoundation/ZEO/blob/afaf0eb9246a1ea25c60164ec75cdd6a47545ad8/src/ZEO/asyncio/client.py#L915-L940

In essence this PR wraps call to Protocol._connecting.cancel() because CanceledError is thrown immedeately from under this call.

In https://github.com/zopefoundation/ZEO/pull/182#issuecomment-826630195 I, being new to asyncio, attributed this "immedeate throw" to the fact that canceled future was Fut, not asyncio. But today, while checking afresh, I've made sure that Protocol._connecting is actually a _full_ asyncio.Future, and that cancelling it should *not* immediately throw CanceledError exception.

So I started to dig into why it happens to be like this, where, despite of intended behaviour that CanceledError will be raised _eventually_ into the awaiter, we have it raised immediately. And what I found is ...

... that ZEO tests, use special event loop implementation, _the incorrectly implements intended asyncio semantic_. For example ZEO.asyncio.testing.Loop.call_soon invokes provided callback immediately :

https://github.com/zopefoundation/ZEO/blob/afaf0eb9246a1ea25c60164ec75cdd6a47545ad8/src/ZEO/asyncio/testing.py#L16-L31

instead of scheduling it to be called later: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_soon

And that is what we are seeing in the failing tracebacks in https://github.com/zopefoundation/ZEO/issues/165 :

```
...
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 447, in registered
    self._clear_protocols(protocol)
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 404, in _clear_protocols
    p.close()
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 98, in close
    self._connecting.cancel()
  File "/home/jens/src/ZEO/src/ZEO/asyncio/testing.py", line 31, in call_soon
    func(*args)                                                                         <-- call_soon
  File "/home/jens/src/ZEO/src/ZEO/asyncio/client.py", line 129, in done_connecting     <-- invokes func
    if future.exception() is not None:                                                  <-- immediately
asyncio.exceptions.CancelledError
```

So this PR adjusts non-test code to work correctly with the particular wrong implementation of ZEO.asyncio.testing.Loop.

BUT

So adjusted code will work *incorrectly* when run in production with standard - non-testing - asyncio event loop. There, that CanceledError will _not_ be raised on Protocol._connecting.cancel() - it will be raised into the awaiter of _connecting future.

Do you see now that **it is pointless to try to bring tests to zero failures with current implementation of ZEO.asyncio.testing.Loop**?

Because while doing so we get false sense of correctness, while in actual production usage the code will have different behaviour.

To me, the first thing to do in this situation is to either fix ZEO.asyncio.testing.Loop, or to understand why we need it at all and replace its usage with standard asyncio event loop.

I'm sorry, I do not want to block progress, but in my view this issue is just too big to put aside or ignore.

Kirill

dataflake:
I get your point about the test results being invalid while the test fixture itself is faulty. I personally have no knowledge about async programming at all so I cannot fix this. Would you know how to fix ZEO.asyncio.testing.Loop?

navytux:
I tried to have a look there. This special Loop is not `.run()` or `.run_forever()` anywhere - asyncio/tests.py call it step by step, and, for example for `loop.call_later` explicitly verify what has been scheduled and then explicitly call scheduled functions *from the test code*, e.g.

Loop.call_later:
https://github.com/zopefoundation/ZEO/blob/afaf0eb9246a1ea25c60164ec75cdd6a47545ad8/src/ZEO/asyncio/testing.py#L77-L80

Check/usage in _test_ code:
https://github.com/zopefoundation/ZEO/blob/afaf0eb9246a1ea25c60164ec75cdd6a47545ad8/src/ZEO/asyncio/tests.py#L419-L423

One way I see to make this Loop to be closer to real event loops is to change its `.call_soon` and `.call_soon_threadsafe` not to invoke provided callback, but to queue it into `Loop.soon` and for the test code to verify content of `loop.soon` similarly to how it verifies `loop.later`. But this means we need to update all test code in asyncio/test.py.

I'm too not an async-person(*), but that what comes to mind after quick inspection.

Hope it helps a bit,
Kirill

(*) I find async programming really messy for no good reason - throwing uninteresting and unneeded details at application developer which make program logic only more twisted and hard to follow, with exposing lots of low-level details that make it really easy to make mistakes. Once again, as said in https://github.com/zopefoundation/ZEO/pull/182#issuecomment-826630195, my view is that gevent/greenlet/pygolang/... provide better approach.

d-maurer:
Kirill Smelkov wrote at 2022-3-18 07:49 -0700:
> ...
>... that ZEO tests, use special event loop implementation, _the incorrectly implements intended asyncio semantic_. For example ZEO.asyncio.testing.Loop.call_soon invokes provided callback immediately :
>
>https://github.com/zopefoundation/ZEO/blob/afaf0eb9246a1ea25c60164ec75cdd6a47545ad8/src/ZEO/asyncio/testing.py#L16-L31

I think you are on the wrong route.

In my view, the problem is at a different place:
future callbacks should never raise an exception -- because those
exceptions arrive at the loop and not application code.

In other words: I believe that the `create_connection` done callback
is buggy:
```
        self._connecting = cr = asyncio.ensure_future(cr, loop=self.loop)

        @cr.add_done_callback
        def done_connecting(future):
            if future.exception() is not None:
                logger.info("Connection to %r failed, retrying, %s",
                            self.addr, future.exception())
                # keep trying
                if not self.closed:
                    self.loop.call_later(
                        self.connect_poll + local_random.random(),
                        self.connect,
                        )
```
`future.exception()` will raise an exception if the future was canceled
(or `create_connection` raised an exception).
Obviously, the code above expects that `future.exception()` returns
exception information, not that is raises an exception.
A fixed callback implementation would be:
```
        def done_connecting(future):
            try: future.exception()
	    except Exception:
                logger.exception("Connection to %r failed", self.addr)
                # keep trying
		...
```

You are right that the test loop behaves differently from
the real loop. But using mockup objects for tests is quite common --
and not a problem on its own.
For a loop controlled externally, it does not make a big
difference if `call_soon` calls immediately or schedules the callback
to be called during the next loop turn -- at least this is the case
for the current situation.

navytux:
> I think you are on the wrong route.
>
> In my view, the problem is at a different place:
> future callbacks should never raise an exception -- because those
> exceptions arrive at the loop and not application code.

@d-maurer, thanks for feedback. Maybe you are right. If future callbacks should not raise exceptions via regular raise, we might be able to try to fix this problem without testing.Loop rework.

navytux:
Note, that Future.exception() _returns_ exception, except it raises it for CanceledError: https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.exception .

d-maurer:
Dieter Maurer wrote at 2022-3-18 19:15 +0100:
> ...
>In other words: I believe that the `create_connection` done callback
>is buggy:
>```
>        self._connecting = cr = asyncio.ensure_future(cr, loop=self.loop)
>
>        @cr.add_done_callback
>        def done_connecting(future):
>            if future.exception() is not None:
>                logger.info("Connection to %r failed, retrying, %s",
>                            self.addr, future.exception())
>                # keep trying
>                if not self.closed:
>                    self.loop.call_later(
>                        self.connect_poll + local_random.random(),
>                        self.connect,
>                        )
>```
>`future.exception()` will raise an exception if the future was canceled
>(or `create_connection` raised an exception).
>Obviously, the code above expects that `future.exception()` returns
>exception information, not that is raises an exception.
>A fixed callback implementation would be:
>```
>        def done_connecting(future):
>            try: future.exception()
>	    except Exception:
>                logger.exception("Connection to %r failed", self.addr)
>                # keep trying
>		...
>```

The last part is wrong -- `future.exception()` does return, not raise
an exception, if the exception was set. It raises an exception
if the future was canceled.

navytux:
@d-maurer, thanks for correcting my analysis. Now I can approve this pull-request.

navytux (approved):

dataflake:
I really appreciate all your work!

d-maurer:
Kirill Smelkov wrote at 2022-3-18 23:23 -0700:
***@***.***, thanks for correcting my analysis. Now I can approve this pull-request.

We might still wait a bit before merging:
I have seen a recent change regarding `ClientDisconnected` handling
(I almost had it discarded by accident).
In addition, I remember (vaguely) a related discussion.
I would like to look into this a bit more:
I have the feeling that there is no need for concern about
`ClientDisconnected`: almost all methods are called via
`call_async_threadsafe`. Exceptions from those methods
will be propagated to the caller. I expect it to handle
`ClientDisconnected` and to properly shut down the connection
potentially with a reconnection trial. This logic is somewhere
but I do not yet know where. If it is at an appropriate place,
we can let it handle all `ClientDisconnected` exceptions and
do not need to worry about it in asynchronous code.

dataflake:
@d-maurer I already merged it, but that's not a problem, I won't make a release until I get feedback from you about that issue. I suggest you create a separate branch/PR.

d-maurer:
Dieter Maurer wrote at 2022-3-19 08:04 +0100:
>Kirill Smelkov wrote at 2022-3-18 23:23 -0700:
***@***.***, thanks for correcting my analysis. Now I can approve this pull-request.
>
>We might still wait a bit before merging:
>I have seen a recent change regarding `ClientDisconnected` handling
>(I almost had it discarded by accident).
>In addition, I remember (vaguely) a related discussion.
>I would like to look into this a bit more:
>I have the feeling that there is no need for concern about
>`ClientDisconnected`: almost all methods are called via
>`call_async_threadsafe`. Exceptions from those methods
>will be propagated to the caller. I expect it to handle
>`ClientDisconnected` and to properly shut down the connection
>potentially with a reconnection trial. This logic is somewhere
>but I do not yet know where. If it is at an appropriate place,
>we can let it handle all `ClientDisconnected` exceptions and
>do not need to worry about it in asynchronous code.

The change was good -- it removed special handling of `ClientDisconnected`
(in `ZEO.asyncio.client.Client.verify`) and handled exceptions
from `server.lastTransaction` (exactly as those from other
server calls in this method) in the correct way: let the registration fail.


Results from my analysis:

 * all server calls can raise `ClientDisconnected` (as well as other
   server exceptions)

 * the reconnection logic works as follows:
   the "protocol" (--> `ZEO.asyncio.client.Protocol`) is informed
   by `asyncio` about the connection loss via a call to
   `connection_lost`.
   This calls `client.disconnected` which tries to reconnect.

navytux:
@d-maurer, thanks for confirming that https://github.com/zopefoundation/ZEO/commit/171c2a60351f5718b6e52efa33a87ddb3aae66fe is correct and for sharing your thoughts.

@dataflake, you are welcome.

PR #186, opened 2022-03-11
fix #150

### LINKED ISSUE #150: Client incorrectly interleaves blob calls on large enough blobs
After #149 I of course wanted to know why the mentioned cronjob causes the zeo server to error with "you said blobStart twice" in the first place.

### Bug

The issue I found (after a metric ton of printf debugging) is that [call_async_iter](https://github.com/zopefoundation/ZEO/blob/master/src/ZEO/asyncio/base.py#L141) currently [always directly writes the first message](https://github.com/zopefoundation/ZEO/blob/master/src/ZEO/asyncio/base.py#L81), even when `self.pause` is active at that moment. Which means that the overall ordering of message calls is **not** preserved.

Im my example, the client wants to store two blobs in sequence, and the first one is large enough, so the following can happen:

1. `storeBlob` for the first blob runs `call_async_iter`, which starts writing messages: blobStart, then blobChunk several times. At some point, the flow control kicks in and calls `pause_writing`. `writeit` puts the remaining blobChunk and blobEnd messages into the `self.output` list.
2. If `storeBlob` is now called for the second blob _before_ `resume_writing` comes back in, the second `blobStart` is written immediately -- before the paused blobEnd had a chance to be written.

### Impact 

In general, this means that the order in which messages go out to the server is currently not guaranteed to be the same as they came in to the client. I _guess_ this does not cause more trouble because a) flow control probably only kicks in for blobs and b) only blobs are that sensitive to the order of operations. But this is only a guess, I'm not overly familiar with the ZEO protocol.

### Bugfix

I don't have a proposal, sorry. My first thought was something like this:
```python
def writeit(data):
    data = iter(data)
    if paused:
        append(data)
        return
    for message in data:
        writelines((pack(">I", len(message)), message))
        if paused:
            append(data)
            break
```
but that doesn't fix the problem (I've tried it).

d-maurer:
Fixes #150.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 5.2.4 (unreleased)
 ------------------
 
+- Fix bug related to blobs stored by ``ZEO``
+  `#150 <https://github.com/zopefoundation/ZEO/issues/150>`_.
+
 
 5.2.3 (2021-08-09)
 ------------------
--- a/src/ZEO/asyncio/base.py
+++ b/src/ZEO/asyncio/base.py
@@ -77,6 +77,9 @@ class Protocol(asyncio.Protocol):
             # will be used with blobs, in which case, the individual
             # messages will be big to begin with.
             data = iter(data)
+            if paused:
+                append(data)
+                return
             for message in data:
                 writelines((pack(">I", len(message)), message))
                 if paused:
--- a/src/ZEO/asyncio/tests.py
+++ b/src/ZEO/asyncio/tests.py
@@ -2,8 +2,12 @@ from .._compat import PY3
 
 if PY3:
     import asyncio
+    def to_byte(i):
+        return bytes([i])
 else:
     import trollius as asyncio
+    def to_byte(b):
+        return b
 
 from zope.testing import setupstack
 from concurrent.futures import Future
@@ -18,6 +22,7 @@ import unittest
 
 from ..Exceptions import ClientDisconnected, ProtocolError
 
+from .base import Protocol
 from .testing import Loop
 from .client import ClientRunner, Fallback
 from .server import new_connection, best_protocol_version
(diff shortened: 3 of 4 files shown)


dataflake (commented):
I don't have the ZEO knowledge to review the code change and never used blob storage, sorry

navytux:
@d-maurer, thanks for asking. In https://github.com/zopefoundation/ZEO/issues/150 @wosc identified the problem that `writeit(data)` always writes first element from `data` even if writing is currently paused. This causes issues because messages are sent to the server not in the order in which the client wanted them to be send.

With your patch, `writeit` becomes

```py
def writeit(data):
    data = iter(data)
    if paused:
        append(data)
        return
    for message in data:
        writelines((pack(">I", len(message)), message))
        if paused: 
            append(data)
            break
```

Which should resolve this particular problem and looks to be generally correct.

Kirill

navytux (approved):

navytux:
(I had not looked at tests, only at the change to `writeit`)

d-maurer:
Kirill Smelkov wrote at 2022-3-11 10:36 -0800:
>(I had not looked at tests, only at the change to `writeit`)

Thank you for your review!

navytux:
Dieter, you are welcome.

icemac:
Triggering CI after reactivating GHA to see that tests pass.

dataflake:
Interesting how it's the same test for dropping versus verifying caches that fails. Weren't there recent changes around that functionality? Or it might be a timing issue.

d-maurer:
`drop_cache_rather_than_verify` is the only failing test -- but apparently it does not always fail. I do not yet have an explanation for this behavior. Anyway, it has nothing to do with blob handling - and thus with this PR.

icemac:
So we are actually still testing no newer Python versions than 3.7?! – What a pity.
I see there is #165, Maybe we find a way to tackle it later on.

dataflake:
> So we are actually still testing no newer Python versions than 3.7?! – What a pity. I see there are #165 and #182, Maybe we find a way to tackle them later on.

#182 was started a year ago and then stalled. The main issue - as always when it comes to anything related to the ZODB - is lack of developers with enough knowledge and willing to invest the time.

d-maurer:
The checks now have all passed -- without a significant change. Thus, the former failures came from  a race condition.

My last commits introduced debugging code (in the failing test) to get hints about the server response causing the failure. Maybe, they allow to understand the race when the failure occurs the next time.

dataflake (approved):
Thanks for the work!

PR #187, opened 2022-03-16
hunt for the drop cache race condition

d-maurer:
This PR adds more debugging code in the hope to understand the race condition in the `drop_cache_rather_than_verity` test. In my local environment, the race does not occur. Thus, the need to use the github tests (where the race sometimes occurs).

### DIFF
--- a/src/ZEO/tests/drop_cache_rather_than_verify.txt
+++ b/src/ZEO/tests/drop_cache_rather_than_verify.txt
@@ -78,7 +78,7 @@ Now, we'll restart the server on the original address:
 ##### debugging only ########
     >>> print(db.storage._server.client.verify_result)
     cache too old, clearing
-    >>> print(len(handler.records) > 1 and handler)
+    >>> print(len(handler.records) > 1 and str(handler) + handler.records[0].exc_text)
     False
 
 Now, let's verify our assertions above:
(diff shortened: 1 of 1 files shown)


d-maurer:
The `StaleCache` event is correctly notified but the event handler set up by the test raises an exception; therefore, it does not record the event -- this lets the test fail.

Unfortunately, the used (test) logging handler apparently does not record traceback information. Therefore, I do not yet know what exception was raised.

d-maurer:
```
      File "/home/runner/work/ZEO/ZEO/.tox/py37-uvloop/lib/python3.7/site-packages/ZEO/asyncio/client.py", line 521, in verify
        ZEO.interfaces.StaleCache(self.client))
      File "<doctest drop_cache_rather_than_verify.txt[20]>", line 4, in event_handler
        len(e.storage._server.client.cache), str(handler), e.__class__.__name__))
    AttributeError: 'ClientStorage' object has no attribute '_server'
```
This explains the race condition: the `StaleCache` event handler accesses the cache via `e.storage._server.client.cache` (rather than `e.storage._cache`). However, the storage gets its `_server` attribute only during the connection. This is, in some cases, apparently too late for the event handler. Let's change the event handler.

d-maurer:
This PR makes the test `drop_cache_rather_than_verity` more robust; formerly, it suffered occasional failures from a race condition.

I do not think that a formal review is necessary. Unless someone objects, I will merge in a few days.

dataflake (approved):
This is great. The ZEO test code is full of these race conditions.

PR #190, opened 2022-03-19
Remove testing for mtacceptor and dead Python versions

### LINKED ISSUE #188: Remove possibly unused MTAcceptor feature
I have a feeling that this package is carrying a lot of baggage in terms of features that are covered by tests but I have never heard anyone use them.

When setting the environment variable `ZEO_MTACCEPTOR` to a true value enables a feature whereby the ZEO storage server runs each connection in its own thread. The comment at the top of the module `asyncio/mtacceptor.py` says:

``
Testing serveral years ago
suggsted that this was a win, but ZODB shootout and another
lower-level tests suggest otherwise.  It's really unclear, which is
why we're keeping this around for now.
``

The code has been retained for years and years and updated where needed. I wonder if it's used anywhere instead of being nothing but a maintenance burden.

dataflake:
Fixes #188 

This PR removes GitHub Actions tests for the `asyncio/mtacceptor` module because it appears to be unused. The module itself remains in place and will be removed in ZEO version 6.

I also cut some other GitHub Actions tests like Python 3.5, it's dead.


d-maurer (approved):

navytux:
@dataflake, thanks for asking. I'm ok for mtacceptor to go. However I have several suggestions:

- Adjust StorageServer.py, so that if $ZEO_MTACCEPTOR is set, it prints/logs corresponding deprecation warning.
- Target mtacceptor removal for ZEO5.4 instead of ZEO6. This point is explained below:

The company I work for still uses ZODB**4** because it is hard for us to upgrade to ZODB5. We are working on it, but it takes time and needs non-small effort. With ZEO6/ZODB6 I feel people might pretend that "it is a new major version, we can break backward compatibility". And I *strongly suggest avoiding doing so*.

ZODB/Zope ecosystem is already not big. And every breaking change would fragment it more and more.

So I would suggest to do the following:

- ZEO5.3 - deprecate mtacceptor, print corresponding warnings that it will be removed in ZEO5.4;
- ZEO5.4 - remove mtacceptor for real;
- Never do ZODB6/ZEO6 - just continue to issue 5.X with ever growing X.

ZODB5 picked up many good architectural changes. But it also brought with it many significant regressions of data-corruption kind. It took us a while to discover and hopefully fix them, and speaking for ZEO5, I'm still not completely sure that everything was found/fixed. Let's please avoid doing that kind of breakage once again.

Hope, I will be heard,

Kirill

dataflake:
@navytux I'm confused by your comment. Removing `mtacceptor` is, by definition, a backwards-incompatible change. With all the bigger packages under the zopefoundation GitHub organization that I am involved in we have the policy that backwards-incompatible changes require a major release.

Secondly, the actual removal is not something that must happen in the near term. ZEO 6 won't happen any time soon. But it must happen at some point because other changes on the horizon that are not backwards-compatible require a major release as well, like officially dropping support for dead Python versions. 

Could it be that you're attaching too much significance to the version number change? ZEO 6 will not automatically bring any critical architectural changes.

dataflake:
> The company I work for still uses ZODB**4** because it is hard for us to upgrade to ZODB5. We are working on it, but it takes time and needs non-small effort. With ZEO6/ZODB6 I feel people might pretend that "it is a new major version, we can break backward compatibility". And I _strongly suggest avoiding doing so_.
> 
> ZODB/Zope ecosystem is already not big. And every breaking change would fragment it more and more.

One final thought from my side: Do you really believe a major version jump will all of a sudden cause major new developments for ZEO? Look at development activity over the last few years. It's just not happening. ZEO is at a mature stage. The few remaining developers fix bugs, ensure compatibility with newer Python versions, or (like in this case) remove unused code to make maintenance easier. No one is even thinking about features or breaking changes.

d-maurer:
Jens Vagelpohl wrote at 2022-3-20 02:20 -0700:
> ...
>One final thought from my side: Do you really believe a major version jump will all of a sudden cause major new developments for ZEO? Look at development activity over the last few years. It's just not happening. ZEO is at a mature stage. The few remaining developers fix bugs, ensure compatibility with newer Python versions, or (like in this case) remove unused code to make maintenance easier. No one is even thinking about features or breaking changes.

I agree with you in priciple.

However, we might reimplement the `asyncio` integration
once Python 2 support has been dropped.
I find the current implementation very difficult to understand
and it takes me several hours for reexploration each time I look into it:
it uses `asyncio` for connection management and IO basics
but implements its own coroutine executer for higher protocol levels
with its own futures. I believe that the use of "async/await"
could make it much more readable and maintainable.

In principle, a reimplemention of the `asyncio` integration
should preserve backward compatibility. But, there is the risk
of the introduction of new bugs. Therefore, a new major version
would be preferable for such a project.

dataflake:
I agree that there are many places that could benefit from reimplementing them. But who will spend that time? To me that's the biggest hurdle. ZEO 5 was released in 2016 and it had breaking changes. Everything since then has been bug fixes and small improvements.

My personal goals for ZEO 6 would just be cleanup: Remove official support for dead Python versions and remove unused code as much as possible. I would target version 7 for bigger changes.

.github/workflows/tests.yml
```diff
@@ -17,28 +17,20 @@ jobs:
         - ["2.7",   "py27-msgpack1"]
         - ["2.7",   "py27-zeo4"]
         - ["2.7",   "py27-zodbmaster"]
-        - ["3.5",   "py35"]
```
icemac, line 20: All other zopefoundation packages which still support Python 2.7 also support 3.5 as Zope 4 supports both of them. Is there another reason for removing Python 3.5 support besides that is old and dead?

d-maurer:
Michael Howitz wrote at 2022-3-22 00:32 -0700:
***@***.*** commented on this pull request.
>> @@ -17,28 +17,20 @@ jobs:
>         - ["2.7",   "py27-msgpack1"]
>         - ["2.7",   "py27-zeo4"]
>         - ["2.7",   "py27-zodbmaster"]
>-        - ["3.5",   "py35"]
>
>All other zopefoundation packages which still support Python 2.7 also support 3.5 as Zope 4 supports both of them. Is there another reason for removing Python 3.5 support besides that is old and dead?

The time to run the GHA checks?

ZEO (at least the test suite) is plagued by race conditions.
Some of them are not observable in local environments and
show up only in GHA checks.

The ZEO GHA workflow treats a large number of variants:
with/without mtacceptor, uvloop, msgpack, ZEO4, zodbmaster.
As a consequence, the GHA checks take hours.

I am happy when rarely used things stop to get testsd; this
reduces the GHA check time and allows a more efficient fix
of race conditions only observed on github.

.github/workflows/tests.yml
dataflake: I am not "removing Python 3.5 support", I am just removing the test for it on GitHub Actions. Those tests burn thousands of useless minutes for a Python version that is EOL.

icemac: Okay, let's stop testing Python 3.5. Even though I cannot get rid of the feeling that not testing it means not longer supporting it: We will not see problems arising with that version as time goes by.

icemac (approved):
LGTM.

.github/workflows/tests.yml
dataflake: Look at it this way: If anyone comes to us with a problem report for Python 3.5 we will definitely help. That's sufficient to claim support.

d-maurer:
Michael Howitz wrote at 2022-3-24 00:17 -0700:
***@***.*** commented on this pull request.
>> @@ -17,28 +17,20 @@ jobs:
>         - ["2.7",   "py27-msgpack1"]
>         - ["2.7",   "py27-zeo4"]
>         - ["2.7",   "py27-zodbmaster"]
>-        - ["3.5",   "py35"]
>
>Okay, let's stop testing Python 3.5. Even though I cannot get rid of the feeling that not testing it means not longer supporting it: We will not see problems arising with that version as time goes by.

Do you see a way to influence which GHA jobs are run?

We could then run full checks during the periodic automatic
checks and limit the number of checks for push.

navytux:
@dataflake, everyone, thanks for feedback. Yes, my main concern is that major version change is perceived by many people as giving the right to make breaking changes.

Removing mtacceptor is not the kind of breakage I'm talking about. Removing mtacceptor is removing likely unused feature with deprecation warning emitted and, I assume so, that if, during deprecation period, someone speaks up that mtacceptor is still useful to that person's use-case, the decision to remove mtacceptor will be reconsidered one way or another. In other words the change to remove mtacceptor is designed not to cause any breakage for real.

On the other hand the kind of breaking change I'm talking about is change in behaviour: change of public API, change of semantic of already established API, etc. Even after Python2 -> Python3 breakage (which, personally, I perceive as disaster), and despite that breakage caused lots of problems worldwide, and it took more than **ten** years for the world to _kind of_ migrate to Python3(*), in my view many still believe that it is a good idea to introduce breaking changes and to _"just name it with a new major version"_ for it to be ok.

In an ideal world the change of version should not matter that much, but in reality the change from ZODB/ZEO 5 to 6 will be/is perceived by many as the valid reason to change what should not be changed while it is all 5.x.

I want to avoid breaking changes for ZODB. That's why I proposed to have a clear signal with such promise.

Kirill

(*) there are still many big projects that run only on Python2 even as of today. Porting non-small codebases, that take their origins in 2000s, from Python2 to Python3 requires significant effort and gives not so much in return. That's why projects are slow to migrate. And that's why e.g. PyPy is not going to drop Python2 support - to still provide support to such codebases.

P.S. I apologize for the delay in replying as last several days were a bit special on my side.

dataflake:
Again, for this particular project I don't see people lining up to make breaking changes, we're happy there still are people with the required knowledge to make any changes at all.

In general, breaking changes will happen. No one can give you a guarantee that they won't happen. What we usually do in the bigger Zope Foundation projects is to create maintenance branches for older versions. If there is demand and if there are people willing to spend the time fixes can be applied to those maintenance branches and releases can be made. So for example as soon as any backwards-incompatible change like the actual removal of mtacceptor or removal of support for old Python version happens, a maintenance branch for the 5.x release will be created and `master` will become ZEO version 6. Even after ZEO 6 is released you could see new ZEO 5 releases coming from that maintenance branch. It's a matter of demand and manpower.

Keep in mind that whenever you're asking "don't break or change X" you are implicitly asking the code maintainers to invest their spare time to continue support for X. That's OK if you'd be willing to do the actual work yourself. But you may see pushback if you expect others to do that work for you.

d-maurer:
Kirill Smelkov wrote at 2022-3-24 06:37 -0700:
> ...
>I want to avoid breaking changes for ZODB. That's why I proposed to have a clear signal with such promise.

You have proposed to drop `loadBefore` :-;

d-maurer:
Kirill Smelkov wrote at 2022-3-24 08:04 -0700:
>> You have proposed to drop `loadBefore` :-;
>
>Via first deprecating it. In my view there is a bit difference in between a) breaking something in one go, and b) having a smooth transition when things continue to work in both old and new ways during transition period.

It would introduce a (quite serious) backward incompatibility
(affecting all storage maintainers).
According to our rules, this would require a new major release.
You do not want any new major release....

navytux:
No, no incompatibility is introduced. Even if a storage implements only loadBefore things will continue to work as before right after the change I propose. Backward compatibility is preserved. So this does _not_ need a new major release in my view.

To sum it up:

- I'm not proposing to stop progress;
- I'm suggesting to pay more attention to backward compatibility;
- I'm pointing out a correlation in between new major version bump and mental tendency to use those as ocassion for introducing breakage "because now we can".

.github/workflows/tests.yml
icemac: Okay, convinced.

icemac:
@d-maurer wrote:
> Do you see a way to influence which GHA jobs are run? We could then run full checks during the periodic automatic checks and limit the number of checks for push.

I think it is possible to ignore configurations depending on the signal which caused the CI run (periodic vs. push). But I currently lack the steam to look it up and try it out.

PR #192, opened 2022-03-30
tpc_vote/abort + log message enhancements

### LINKED ISSUE #156: "Bad request, 'vote'"
In a ZEO (5.2.1) logfile, I see several `Bad request, 'vote'` messages.

In all those cases, I see the following sequence of log entries:
 * the error information (including a traceback) for a failed conflict resolution
 * `Bad request, 'vote'` repeating the previous traceback, now associated with a `ConflictError`.
 * `no current transaction: tpc_abort()`

Apparently, ZEO turns a `ConflictError` into a `Bad request: 'vote'` (which in my view is wrong: it simply indicates that `vote` answers "no") and after that it gets confused by a `tpc_abort`.

d-maurer:
Fixes #156.

This PR
* adds `ConflictError` to the list of exceptions not logged by the server: this exception is expected to be raised by server operations; the client/its application should decide whether it wants it logged.
* suppress server log message `no current transaction: tpc_abort()`. The storage API allows `tpc_abort` to be called with an invalid transaction (the call should be ignored in this case); the server's `tpc_vote` implementation invalidates the transaction in case of errors and thereby relies on a call of `tpc_abort` with an invalid transaction.
* changes the label for log messages reporting exceptions in server calls from `Bad request ...` to `... raised exception:`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,11 +4,27 @@ Changelog
 5.3.1 (unreleased)
 ------------------
 
+- Add ``ConflictError`` to the list of unlogged server exceptions
+  (the client/its application should determine whether it wants
+  them logged).
+
+  Prevent ``no current transaction: tpc_abort()`` server log entries.
+  The storage API allows ``tpc_abort`` to be called with an
+  invalid transaction (the call should be ignored in this case)
+  and the server's ``tpc_vote`` relies on this.
+
+  Change the server's log message label for request exceptions
+  from ``Bad request ...`` to ``... raised exception:``,
+  hinting towards a server rather than client problem.
+
+  See `issue 156 <https://github.com/zopefoundation/ZEO/issues/156>`_.
+
+
 
 5.3.0 (2022-03-24)
 ------------------
 
-- Remove tests for the `asyncio/mtacceptor` module, it appears unused
+- Remove tests for the ``asyncio/mtacceptor`` module, it appears unused
   and presents a maintenance burden. The module will be removed in
   ZEO version 6.
 
--- a/src/ZEO/StorageServer.py
+++ b/src/ZEO/StorageServer.py
@@ -182,17 +182,22 @@ class ZEOStorage(object):
             raise ReadOnlyError()
         if self.transaction is None:
             caller = sys._getframe().f_back.f_code.co_name
-            self.log("no current transaction: %s()" % caller,
-                     level=logging.WARNING)
+            # ``tpc_abort`` is allowed to be called with invalid transaction
+            # ``vote`` relies on this
+            if caller != "tpc_abort":
+                self.log("no current transaction: %s()" % caller,
+                         level=logging.WARNING)
             if exc is not None:
                 raise exc(None, tid)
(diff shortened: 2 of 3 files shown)


dataflake (approved):

navytux:
@d-maurer, thanks for asking for review. I understand I'm late to the party, but here it is anyway:

- I agree that *ConflictError should not be logged on server, as conflicts are one of possible expected normal outcome when a client tries to commit.
- `changes the label for log messages` -> this looks reasonable

about the following point (emphasis mine):

> - suppress server log message no current transaction: tpc_abort(). The storage API allows tpc_abort to be called with an invalid transaction (the call should be ignored in this case); the server's tpc_vote implementation invalidates the transaction in case of errors and **thereby relies on a call of tpc_abort with an invalid transaction**.


I believe you are referring to

```py
    def tpc_abort(transaction):
        """Abort the transaction.
        ...
        This call is ignored is the storage is not participating in
        two-phase commit or if the given transaction is not the same
        as the transaction the storage is committing.
        """
```

To me this really looks suspicious and not a very good thing. In my view an invalid transaction parameter passed to a storage method should be considered a bug in API usage. I thus suggest to reconsider removing this warning and also to try to rework part(s) of the code that relies on this.

I understand that this semantic is already established since at least 2007 (zopefoundation/ZODB@c3ce017fb67d3f03ba2c61f1734cd6892e93bf5b), and so it is not practically easy to fix it immediately. Still I would suggest to keep in mind an intent to fix it.

Kirill

d-maurer:
Kirill Smelkov wrote at 2022-4-8 12:02 -0700:
> ...
>I believe you are referring to
>
>```py
>    def tpc_abort(transaction):
>        """Abort the transaction.
>        ...
>        This call is ignored is the storage is not participating in
>        two-phase commit or if the given transaction is not the same
>        as the transaction the storage is committing.
>        """
>```

Indeed.

>To me this really looks suspicious and not a very good thing. In my view an invalid transaction parameter passed to a storage method should be considered a bug in API usage. I thus suggest to reconsider removing this warning and also to try to rework part(s) of the code that relies on this.

The rationale behind ignoring the call if the storage
does not know a transaction:
you should always be able to call `tpc_abort` independently of what
may or may not have succeeded in the past.
Usually, `tpc_begin` starts a transaction (for the ZEO server, it is
actually `tpc_vote` - but the argument is the same).
Assume the call starting the transaction times out -- then you
do not know the storage state: it may or may not have started the
transaction. The grant above allows to call `tcp_abort` to
bring the storage again in a known state whether or not the
previous transaction start call has been successful.

I agree with you that ignoring the call if the transactions
are different is suspicious.
Following the reasoning above, I would expect that
`tcp_abort` aborts the transaction in all cases,
whether or not the transaction information of caller and callee
agree or not; if both assume an active transaction and they
are different, then an error log message would be adequate
(but `tcp_abort` should nevertheless succeed).

navytux:
@d-maurer, thanks for clarifying. What you say is reasonable.

PR #199, opened 2022-04-09
Reduce ZEO testing time by dropping tests with some storages

### LINKED ISSUE #198: Should we stop testing ZEO against all ZODB storage types?
Running the ZEO test suite in a single variant takes (on my local system) between 7 (py3.x) and 10 (pypy 3.x) minutes. I find this quite long.

A major cause for the long time: the tests check ZEO for all ZODB storage types, e.g. for `FileStorage`, for `HexFileStorage`, for `MappingStorage`, for `DemoStorage`, ...
I doubt that this is necessary: we are testing ZEO, not the storages. If ZEO is tested against a model storage, ZEO should work too with other well behaving storages.
In my view, it should be sufficient to test ZEO with `FileStorage` with and without blobs.

Opinions?

d-maurer:
Fixes #198.

The ZEO tests have the task to verify proper ZEO functioning, not to test storages. If ZEO works correctly and a storage works correctly, then ZEO should work correctly with this storage.

This PR drops the ZEO tests with the storages `DemoStorage`, `FileStorage` without blobs, `HexFileStorage` and `ClientHexFileStorage`. It retains the tests with `Filestorage` with shared or server side blobs and `MappingStorage`.

The `MappingStorage` tests cover the case "storage without blobs". It is used instead of `FileStorage` without blobs to increase the chance to detect race conditions (`MappingStorage` has slightly different latency from `FileStorage` because the data is stored in RAM rather than on disk).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,18 @@ Changelog
 5.4.0 (unreleased)
 ------------------
 
+- Test ZEO only with the following storages
+  ``FileStorage`` with server side blobs,
+  ``FileStorage`` with shared blob directory,
+  ``MappingStorage`` (without blobs).
+
+  Those tests cover all storage features with impact on ZEO:
+  without blobs, with shared blobs and with server side blobs;
+  load, store, two phase commit, undo.
+  Therefore, passing tests for those storages provide high confidence that ZEO
+  works for other properly implemented storages as well.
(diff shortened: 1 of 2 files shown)


dataflake (approved):
In my own 6 core test rig where I use `tox -pall` to run all tests defined at the top of the tox configurations concurrently I am seeing 6-8 minutes total instead of 8-10.

PR #202, opened 2022-05-24
*: Documentation, Cosmetics

navytux:
--------
kirr:

Extract from https://github.com/zopefoundation/ZEO/pull/195 bits that
add documentation to existing code without changing semantic, and fix
typos.

The only things I added myself are:

- documentation for server_sync in ClientStorage;
- stub documentation for credentials in ClientStorage.

Even though we agree to deprecate credentials in favour of peer-to-peer
TLS, removing their support should go as a separate step.

For server_sync feature, that
https://github.com/zopefoundation/ZEO/pull/195 currently removes, we
actually do use it for correctness:

https://lab.nexedi.com/nexedi/erp5/blob/eaae74a082a0/product/ERP5Type/tests/custom_zodb.py#L175-179
https://lab.nexedi.com/nexedi/erp5/commit/c663257fa477
https://github.com/zopefoundation/ZODB/commit/9821696f584f

So document it with the intent to preserve it.


d-maurer:
Kirill Smelkov wrote at 2022-5-24 03:18 -0700:
> ...
>For server_sync feature, that
>https://github.com/zopefoundation/ZEO/pull/195 currently removes, we
>actually do use it for correctness:

It should not be necessary -- as the IO thread keeps client and
server in sync.

If you need it, we should at least remove the
`lasttransaction/ping` switch (the switching introduces a cyclic
reference) - necessary only for ZEO 4 support.

navytux:
> > For server_sync feature, that
#195 currently removes, we
actually do use it for correctness:

> It should not be necessary -- as the IO thread keeps client and server in sync.

@d-maurer, thanks for feedback.

Going round-trip to server on sync is needed for correctness to assuredly get database view with ZODB changes that another ZODB client could have done _before_ the `transaction.begin()` call was made. Please see https://lab.nexedi.com/nexedi/erp5/commit/c663257fa477 for detailed description of one case where it is necessarry.

For the reference: due to exactly this reason ZODB/go defaults to `server_sync=y` for correctness, and only allows to opt-in to `server_sync=n` if strict multi-node ZODB synchronization is not required:

https://pkg.go.dev/lab.nexedi.com/kirr/neo/go/zodb#ConnOptions
https://lab.nexedi.com/kirr/neo/blob/789c9ed911ac/go/zodb/db.go#L362-402
https://lab.nexedi.com/kirr/neo/blob/789c9ed911ac/go/zodb/zodb.go#L322-339
https://lab.nexedi.com/kirr/neo/commit/151d8b79bb1a

Does it clarify why `storage_sync` is needed?

navytux:
( thanks @dataflake )

d-maurer:
Kirill Smelkov wrote at 2022-5-24 03:18 -0700:
> ...
>- stub documentation for credentials in ClientStorage.

A comment suggests that
`credentials` work only for ZEO 4 servers;
for ZEO 5, it has been dropped in favor or SSL.

navytux:
@d-maurer, are you satisfied with my explanation about why `storage_sync` is needed, and with `credentials` amendment?

I would like to proceed reviewing https://github.com/zopefoundation/ZEO/pull/195 and it would be good to see whether we agree on hereby step first.

Thanks beforehand for feedback,
Kirill

src/ZEO/ClientStorage.py
```diff
@@ -222,7 +222,8 @@ def __init__(self, addr, storage='1', cache_size=20 * MB,
         username
         password
         realm
-            Credentials for authentication to server.
+            [ZEO4 only] Credentials for authentication to server.
+            In ZEO5 support for credentials has been dropped in favor or SSL.
```
jugmac00, line 226:
```suggestion
            In ZEO5 support for credentials has been dropped in favor of SSL.
```

d-maurer:
Kirill Smelkov wrote at 2022-5-29 03:33 -0700:
***@***.***, are you satisfied with my explanation about why `storage_sync` is needed, and with `credentials` amendment?

I think we have already agreed that ZEO 4 server support gets discontinued:
this would mean that we no longer need `credentials`.

The link you provided for `server_sync` seems to suggest that
this is a feature required when subsystems other than a (single) ZODB
are incorporated.
I recommend to outline in the description situations for which
setting `server_sync` to `True` can be important.
I, at least, would not understand the need from the current description.

src/ZEO/ClientStorage.py
navytux: @jugmac00, thanks for the catch. I'm incorporating your change in my next update.

navytux:
@d-maurer thanks for feedback.

I've tried to add documentation examples where `server_sync` might be needed in https://github.com/navytux/ZEO/commit/87e4c9fbfa7e277cd0feef4501df408136295cb5. Does it look better now?

Regarding `credentials` - yes, we indeed agreed that we'll drop them, but here in this patch I'm doing only one thing - documenting current state. I did not put in a lot of text and documented credentials with only a small stub text, indeed keeping in mind the intent that we'll remove it later in another step - that removes credentials support completely.

Hope it is ok.

d-maurer:
Kirill Smelkov wrote at 2022-5-30 01:23 -0700:
>I've tried to add documentation examples where `server_sync` might be needed in https://github.com/navytux/ZEO/commit/87e4c9fbfa7e277cd0feef4501df408136295cb5. Does it look better now?

Could you please directly work in `zopefoundation/ZEO`
rather than a private fork?

Apparently, I cannot work as usual in your private fork:
e.g. make a suggestion for a range of lines.
In addition it would avoid that I get 2 (rather than 1)
messages about your activities.


I commented on your changes:
convincing examples - but the description could be more concise.


>Regarding `credentials` - yes, we indeed agreed that we'll drop them, but here in this patch I'm doing only one thing - documenting current state. I did not put in a lot of text and documented credentials with only a small stub text, indeed keeping in mind the intent that we'll remove it later in another step - that removes credentials support completely.

Could you add a deprecation notice:
`credential` support is scheduled to be removed in `ZEO6`.

navytux:
@d-maurer, thanks for feedback. I've incorporated your suggestions in https://github.com/zopefoundation/ZEO/commit/ba49bdbf3bea. I've pushed it to both my fork and zopefouindation/ZEO for now, because I cannot change source branch of existing pull request.

I will try to use upstream ZEO repository for new pull requests to match your preference.

Is it all ok with the text now?

d-maurer:
Kirill Smelkov wrote at 2022-5-30 02:59 -0700:
> ...
>Is it all ok with the text now?

You did not follow my first suggestion (regarding the second
paragraph of the `server_sync` doc).

My suggestion contained an explanation where the syncing
actually happens and would allow a more smooth description:

My suggestion (replacement for the second paragraph):

   The `sync` is called in `transaction.begin`.
   A server round trip at this place guarantees that
   the transaction takes notice of all prior modifications.

   This may be important when several client processes
   share the same ZODB as the following examples demonstrate.


Suggestion: add "Example 1" to first example


I would drop the pictures: the descriptions are sufficiently
clear that a picture may not be necessary.


Suggestion: start second example with:

  Example 2: Assume 2 client processes communicate ...



I noticed that `Zope processes` should get replaced by
`(ZEO) client processes`: the examples are valid
for all types of client processes not only Zope processes.

navytux:
Thanks, @d-maurer.

I did not saw your replacement for the second paragraph before. Maybe I have missed it - then I'm sorry. Anyway I've replaced the second paragraph with what you suggest, removed pictures and added "Example X." prefixes.

Regarding whether to use `Zope` or `ZEO client processes`, I still believe that `Zope` is more appropriate for the _example_, because it is more concrete. My idea is that we first provide generic definition, that is general and correct in general case, and then illustrate that definition with more specific examples. To me, for the illustration, "Zope" serves better, than "ZEO client process", because in the examples we are also talking about handling user requests. It is straightforward to see what kind of handling it is if ZODB client process is "Zope". It is less straightforward to see what kind of handling it is for "ZEO client process". With the latter the text would be still correct, but having lesser illustration strenght. Thus I suggest to leave "Zopes" in the examples.

Kirill

d-maurer:
Kirill Smelkov wrote at 2022-5-30 05:00 -0700:
> ...
>Regarding whether to use `Zope` or `ZEO client processes`, I still believe that `Zope` is more appropriate for the _example_, because it is more concrete. My idea is that we first provide generic definition, that is general and correct in general case, and then illustrate that definition with more specific examples.

Ok

d-maurer:
Kirill Smelkov wrote at 2022-5-30 06:25 -0700:
>Thanks. So are we now ok with this patch?

I did not look at the current state but I trust you.
Go ahead.

navytux:
Thanks, Dieter.

navytux:
I've applied the patch to master as https://github.com/zopefoundation/ZEO/commit/d4805a0f0e36e935569a30de06634d4dbb0907b0.

PR #203, opened 2022-05-24
*: Introduce and use NoBlobFileError for situations when blob file is not found

navytux:

### DIFF
--- a/src/ZEO/ClientStorage.py
+++ b/src/ZEO/ClientStorage.py
@@ -670,8 +670,7 @@ class ClientStorage(ZODB.ConflictResolution.ConflictResolvingStorage):
             else:
                 # We're using a server shared cache.  If the file isn't
                 # here, it's not anywhere.
-                raise POSException.POSKeyError(
-                        "No blob file at %s" % blob_filename, oid, serial)
+                raise NoBlobFileError(oid, serial, blob_filename)
 
         if os.path.exists(blob_filename):
             return _accessed(blob_filename)
@@ -702,7 +701,7 @@ class ClientStorage(ZODB.ConflictResolution.ConflictResolvingStorage):
             if os.path.exists(blob_filename):
                 return _accessed(blob_filename)
 
-            raise POSException.POSKeyError("No blob file", oid, serial)
+            raise NoBlobFileError(oid, serial)
 
         finally:
             lock.close()
@@ -726,10 +725,10 @@ class ClientStorage(ZODB.ConflictResolution.ConflictResolvingStorage):
                 if self.shared_blob_dir:
                     # We're using a server shared cache.  If the file isn't
                     # here, it's not anywhere.
-                    raise POSException.POSKeyError("No blob file", oid, serial)
+                    raise NoBlobFileError(oid, serial)
                 self._call('sendBlob', oid, serial)
                 if not os.path.exists(blob_filename):
-                    raise POSException.POSKeyError("No blob file", oid, serial)
+                    raise NoBlobFileError(oid, serial)
 
             _accessed(blob_filename)
             if blob is None:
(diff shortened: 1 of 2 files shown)


dataflake (approved):

navytux:

navytux:
@d-maurer, I hope this patch is also ok. I will wait for either your approval, or another kind of feedback as explained in https://github.com/zopefoundation/ZEO/pull/204#issuecomment-1141331194.

d-maurer (commented):
We likely need also corresponding entries in `exc_classes`, `exc_factories` and (maybe) `unlogged_exceptions` of `ZEO.asyncio.client`. Without those entries, the exception is likely mapped to `ServerException`, potentially with information loss.

navytux:
@d-maurer, thanks, good catch.

For this patch I took everything related to `NoBlobFileError` from https://github.com/zopefoundation/ZEO/pull/195 as is, but as you point out it might be not enough.

From one side the exceptions that `ZEO.ClientStorage`  raises by itself are already originating on client and does not go through decoding as exceptions received via "from-server" messages do. On the other hand it is valid ZEO configuration where ZEO1 is configured to use ZEO2 as its server and it is only ZEO2 that talks to actual storage backend. Thus here, if ZEO2 raises `NoBlobFileError` - or any other error defined in e.g. ZEO.Exceptions - it won't be decoded on ZEO1 and, as you say, will instead be represented via `ServerException`.

The other point to consider is that till now the whole set of exceptions, that ZEO is conveying to its client as is, is only `ZODB.POSException.*` and builtin `TypeError`/`KeyError`:

https://github.com/zopefoundation/ZEO/blob/f64a44ee3b03246723b502e4066d03b4fb8e8252/src/ZEO/asyncio/client.py#L356-L382

So if we add **ZEO-specific** `ZEO.ClientStorage.NoBlobFileError` to that set, it will break that invariant.

And due to the above, now, I'm actually not sure whether hereby patch makes sense as is: if we are going to indicate _"no blob file present"_ via an error, that `IStorage.loadBlob` might raise, the more proper place for such exception to be defined is in ZODB - in `ZODB.POSException` - not in ZEO. Then, if we indeed do it that way, it will be ok to adjust `exc_classes` and friends, and to use that exception in ZEO.ClientStorage.loadBlob implementation.

For now I would suggest not to introduce another error class and just use a internal helper function to create POSKeyError with proper message for "no blob file". Or, alternatively, to completely drop hereby change.

What do you think?

d-maurer:
Kirill Smelkov wrote at 2022-5-31 01:39 -0700:
>So if we add **ZEO-specific** `ZEO.ClientStorage.NoBlobFileError` to that set, it will break that invariant.

I turned a `POSKeyError`, i.e. one of those exceptions specially
handled, into a `NoBlobFileError` (because at that place
`POSKeyError` got unexpected arguments: the first argument was
not `oid`).

Thus, `NoBlobFileError` (a subclass of `POSKeyError` should not
make a problem.

navytux:
@d-maurer, sorry, maybe I'm missing something, but do you follow my argument that by rasing `NoBlobFileError` instead of `POSKeyError`, even if NoBlobFileError inherits from the latter, we are introducing new interface for detecting "blob file not found". So if we do that, that interface should be defined universally in ZODB, not in ZEO, right?

Actually ZODB explicitly says that it will be POSKeyError raised, if loadBlob cannot file blob file:

https://github.com/zopefoundation/ZODB/blob/63dcee8d3bf0df6a05b75a378292a1a6061ce0c2/src/ZODB/interfaces.py#L1398-L1404

I think it should be a good idea to differenciate blob problems from "no such oid" error. But if we do that, the right place to introduce NoBlobFileError is in **ZODB**, not ZEO, because if we add it to ZEO, and users start to depend on it, how would they detect the same problems with e.g. FIleStorage? By catching also another class - e.g. ZODB.Exception.NoBlobFileError, and similarly for other storages?

So to me the right way to introduce NoBlobFileError is via changing interfaces in ZODB. But if changing interfaces is not so easy, or not desirable during this work, we can workaround it via raising POSKeyError just properly prepared via ZEO-specific internal function.

navytux:
Closing as not actual anymore. See https://github.com/zopefoundation/ZEO/pull/195#issuecomment-1374901652 for update.

PR #204, opened 2022-05-24
*: Do not use logging directly to emit log messages

navytux:
--------
kirr:

Always use appropriate logger object set up in a module. It was only
ZEO/tests/ConnectionTests.py which was using logging directly even
though corresponding logger object was set up in that module.

### DIFF
--- a/src/ZEO/tests/ConnectionTests.py
+++ b/src/ZEO/tests/ConnectionTests.py
@@ -80,7 +80,7 @@ class CommonSetupTearDown(StorageTestBase):
         for i in 1, 2, ...
         """
         self.__super_setUp()
-        logging.info("setUp() %s", self.id())
+        logger.info("setUp() %s", self.id())
         self.file = 'storage_conf'
         self._servers = []
         self.caches = []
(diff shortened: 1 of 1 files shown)


dataflake (approved):
I think you created the `5` branch too early. Several of those smaller PRs should go onto `master` before the `5` branch is split off.

navytux:
@dataflake, right. I was going to push those patches to both master and 5, but, it seems, for now, as you say, it is better to undo "5" branch at all. Undone. Thanks for the approval.

navytux:
@d-maurer, are you ok with this patch?

Here, it should probably more straightforward, because there is no questionable things like `server_sync` deletion originally noticed in https://github.com/zopefoundation/ZEO/pull/202. In other words this patch is straightforward extract from https://github.com/zopefoundation/ZEO/pull/195.

navytux:
( if it is too much overhead for you to review such straightforward steps, I can go on my own without requesting your approval )

d-maurer (approved):

d-maurer:
Kirill Smelkov wrote at 2022-5-30 09:29 -0700:
>( if it is too much overhead for you to review such straightforward steps, I can go on my own without requesting your approval )

Indeed, I am not that interested in such simple changes.

navytux:
Thanks, @d-maurer, got it.

PR #207, opened 2022-08-01
Fix lint tests.

icemac:
Fixing the tests broken in https://github.com/zopefoundation/ZEO/runs/7610839197?check_suite_focus=true

### DIFF
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -1005,7 +1005,9 @@ class ClientThread(ClientRunner):
 
 
 class Fut(object):
-    """Lightweight future that calls it's callbacks immediately rather than soon
+    """Lightweight future that calls it's callbacks immediately ...
+
+    rather than soon.
     """
 
     def __init__(self):
--- a/src/ZEO/tests/ZEO4/zrpc/smac.py
+++ b/src/ZEO/tests/ZEO4/zrpc/smac.py
@@ -286,7 +286,7 @@ class SizedMessageAsyncConnection(asyncore.dispatcher):
                     except StopIteration:
                         messages.pop(0)
                     else:
-                        assert(isinstance(message, six.binary_type))
+                        assert isinstance(message, six.binary_type)
                         size += self.__message_output(message, output)
 
             v = b"".join(output)
--- a/src/ZEO/tests/testZEO.py
+++ b/src/ZEO/tests/testZEO.py
@@ -730,8 +730,8 @@ class CommonBlobTests(object):
         filename = self._storage.loadBlob(oid, serial)
         with open(filename, 'rb') as f:
             self.assertEqual(somedata, f.read())
-        self.assertTrue(not(os.stat(filename).st_mode & stat.S_IWRITE))
-        self.assertTrue((os.stat(filename).st_mode & stat.S_IREAD))
+        self.assertTrue(not (os.stat(filename).st_mode & stat.S_IWRITE))
+        self.assertTrue(os.stat(filename).st_mode & stat.S_IREAD)
 
     def checkTemporaryDirectory(self):
         self.assertEqual(os.path.join(self.blob_cache_dir, 'tmp'),


dataflake (approved):

icemac:

navytux:
( @icemac, thanks for asking; I'm ok with this PR as well )

navytux:
( @icemac, when I first looked at this PR yesterday, I remember seeing py27-zeo4 failure, but today I cannot find it. Do you know how to retrieve this old test result and why today it is all green? This failure should not be related to hereby patch, but for me it is important to check the details of it. Thanks beforehand )

navytux:
Ok, I found it: https://github.com/zopefoundation/ZEO/runs/7610968213?check_suite_focus=true .

@d-maurer, FYI, current ZEO5 master failed in `check_race_load_vs_external_invalidate` wrt ZEO4 server as:

```
Failure in test check_race_load_vs_external_invalidate (ZEO.tests.testZEO.BlobAdaptedFileStorageTests)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-zeo4/lib/python2.7/site-packages/ZODB/tests/util.py", line 400, in _
    return f(*argv, **kw)
  File "/home/runner/work/ZEO/ZEO/.tox/py27-zeo4/lib/python2.7/site-packages/ZODB/tests/BasicStorage.py", line 630, in check_race_load_vs_external_invalidate
    self.fail([_ for _ in failure if _])
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 410, in fail
    raise self.failureException(msg)
AssertionError: ['T7: obj1.value (2)  !=  obj2.value (1)']
```

I believe it is important to understand this data corruption, because I've also saw it a couple of times with you changes from #195 as well.

d-maurer:
Kirill Smelkov wrote at 2022-8-2 02:43 -0700:
>Ok, I found it: https://github.com/zopefoundation/ZEO/runs/7610968213?check_suite_focus=true .
>
***@***.***, FYI, current ZEO5 master failed in `check_race_load_vs_external_invalidate` wrt ZEO4 server as:
> ...
>I believe it is important to understand this data corruption, because I've also saw it a couple of times with you changes from #195 as well.

Did we not decide to drop the use case `ZEO 6 client, ZEO 4 server`?
Have you seen corresponding failures involving a ZEO5/6 server?

The ZEO 4 implementation is old by now; I would not like to
invest time to hunt for occasional problems in it.
However, if you have evidence that a similar problem also exisits
with ZEO5/6, I am willing to investigate.

navytux:
@d-maurer, thanks for feedback. Yes, we have decided to drop ZEO6 client / ZEO 4 server, but this failure I've seen here is about ZEO5 client / ZEO4 server. In other words something might be wrong with existing ZEO5 client and the bug might silently get into ZEO6 client as well. But no - so far I have not seen this failure on ZEO5/5 or ZEO6/5, but I had run tests on github only few times.

I understand that ZEO4 might be considered old by now. I just wanted to point out that we have this issue already with existing ZEO5 and that if we see it also later with my patches extracted from #195, the patches are not the cause for this corruptions. For myself, I believe it is also important to understand this issue - whether the bug is indeed in ZEO4 server (and will be gone when ZEO4 server usage is gone), or contrary to that it is somehow related to ZEO5 client or the way ZEO5 client interacts with ZEO4 server.

icemac:
@navytux wrote:
> ( @icemac, when I first looked at this PR yesterday, I remember seeing py27-zeo4 failure, but today I cannot find it. Do you know how to retrieve this old test result and why today it is all green? This failure should not be related to hereby patch, but for me it is important to check the details of it. Thanks beforehand )

It seems to be a flaky test, rerunning the GitHub Actions job made it go away.

d-maurer:
Michael Howitz wrote at 2022-8-2 23:05 -0700:
***@***.*** wrote:
>> ( @icemac, when I first looked at this PR yesterday, I remember seeing py27-zeo4 failure, but today I cannot find it. Do you know how to retrieve this old test result and why today it is all green? This failure should not be related to hereby patch, but for me it is important to check the details of it. Thanks beforehand )
>
>It seems to be a flaky test, rerunning the GitHub Actions job made it go away.

The test is specially designed to check for a race condition.
For race conditions, it is natural that they do not occur deterministically
but only in very special situations.

I am convinced that the test failure has revealed a real problem.
The only question is whether we are interested to hunt for its cause
taking into account that for the moment the problem has only
been observed in the rarely used scenario "ZEO5/6 client + ZEO4 server".

navytux:
@icemac thanks for feedback, and @d-maurer, thanks for clarifying it. Indeed this test is not flaky and is specifically designed to probabilistically catch races related to data corruption. When we were adding this tests we tried to make the probability of revealing issues to be high, and it was indeed high - close to 1 - on the failures that were observed before. But with ZEO5 client - ZEO4 server it seems not to be the case.

An interested reader can check more context about this data corruption bugs here:

https://github.com/zopefoundation/ZEO/issues/155
https://github.com/zopefoundation/ZEO/pull/169
https://github.com/zopefoundation/ZODB/pull/345
https://github.com/zopefoundation/ZEO/issues/166
https://github.com/zopefoundation/ZODB/issues/290
...


Anyway, as Dieter says, what needs to be understood first is whether this failure revealed a bug in ZEO5 client, that can also happen when that client operates wrt ZEO5 server, or it was conseqence of how ZEO 5 client particularly interoperates with ZEO4 server.

If it was the latter, it might be not so important, because we already agreed to drop ability for new ZEO client to be able interoperate with older ZEO server. I believe no one is using this mode anyway.

But if it was the former - it might be that not all data corruption bug were fleshed out from ZEO5 yet. And that why (at least for me) it is important to understand what was it.

For the reference where I work we are still using ZODB4 stack (either pristine, or with [MVCC backport from ZODB5](https://lab.nexedi.com/nexedi/ZODB/commit/8e7eab330154?expanded=1)), and we no longer see such data corruption failures for a long time.

d-maurer:
Kirill Smelkov wrote at 2022-8-3 01:30 -0700:
> ...
>But if it was the former - it might be that not all data corruption bug were fleshed out from ZEO5 yet. And that why (at least for me) it is important to understand what was it.

Under Python 3, thread switching is quite lazy (every 5 ms by default).
This means that race conditions can remain hidden easily.

When I remember right, you have implemented an
instrumentation to run a test with much higher
thread switching rate, increasing the probability to detect
a race condition.
If not yet done, the test we are speaking about should get
instrumented in this way. With a bit of luck, the problem
could then be observed much more often and maybe also in
differerent contexts (e.g. ZEO5 client/ZEO5 server).

navytux:
@d-maurer, thanks. Yes, we should already have such instrumentation on this test based on your earlier suggestion:

https://github.com/zopefoundation/ZODB/commit/ef9e02df5097

d-maurer:
Kirill Smelkov wrote at 2022-8-3 02:07 -0700:
***@***.***, thanks. Yes, we should already have such instrumentation on this test based on your earlier suggestion:

The ZEO 4 server behaves differently (than a ZEO 5 server)
when a connection is opened (i.e. in the test failure situation).
While the ZEO 5 server sends its current transaction id
as response to the `register` call, a ZEO 4 server requires
a separate `lastTransaction` call to provide this information.
The server will start sending invalations as soon as it has
answered the register call.
As a consequence with a ZEO 4 server, the client can receive
invalidations earlier than its `lastTid`; with a ZEO 5 server
this will not be the case.

Maybe, the test failure is related to this
special behavior of the ZEO5 client/ZEO4 server combination
during connection establishment.

navytux:
@d-maurer, thanks for feedback. It is indeed like you describe and ZEO5 client even have a special "ZEO4 workaround" to queue invalidations until first `lastTransaction` RPC is complete:

https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/asyncio/client.py#L437-L441
https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/asyncio/client.py#L761-L770
https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/asyncio/client.py#L595-L605

However this particular pecularity is in effect only during and right after a ZEO client established connection to ZEO server.

The failure we saw here is indeed close to such start:

> AssertionError: ['T7: obj1.value (2)  !=  obj2.value (1)']

It was on second iteration where `obj1` was reloaded from storage afresh and `obj2` was retrieved from `ZODB.Connection` cache. That latter cache missed the invalidation for second transaction that changed both `obj1` and `obj2`.

However in https://github.com/zopefoundation/ZEO/runs/7628180513?check_suite_focus=true I also saw the following:

> AssertionError: ['T5: obj1.value (25)  !=  obj2.value (26)']

which was far away from ZEO connection establishment and means that either a) ZODB.Connection cache was wrongly updated to contain future value coresponding to current zconn.at, or b) cold loading from the storage (but still using ZEO client cache) returned wrong result missing the invalidation for txn26.

How do you think, is it that both of this scenarious could be explained by ZEO5/ZEO4 connection establishment pecularity?

d-maurer:
Kirill Smelkov wrote at 2022-8-3 03:53 -0700:
> ...
>How do you think, is it that both of this scenarious could be explained by ZEO5/ZEO4 connection establishment pecularity?

I only looked of the start of
"https://github.com/zopefoundation/ZODB/commit/ef9e02df5097",
saw there `check_race_loadopen_vs_local_invalidate`
and wrongly concluded that this were the failing test.
But you have reported a test failure for
`check_race_load_vs_external_invalidate`.
For this test, the behavior of connection establishment
(and therefore the special lastTid` handling by the ZEO 4 server)
should not be relevant.


> However in https://github.com/zopefoundation/ZEO/runs/7628180513?check_suite_focus=true I also saw the following:
>
> > AssertionError: ['T5: obj1.value (25)  !=  obj2.value (26)']

This observation is particularly strange:
`obj1` contains the value immediately read from the storage
(avoiding the connection (aka pickle) cache);
the value of `obj2` is likely from the connection cache.
That the value of `obj1` is smaller than that of `obj2` indicates
that it is older. Thus, we have read older data from the storage itself
than from the connection cache.

In the test, each thread uses its own ZODB with its own client storage.
Each thread effectively uses the same ZODB connection for each check round
which implies the same connection cache.
The newer value (`26`) must have been stored previosly in the cache.
There are 2 possibilities:
 1. the thread itself has set this value
 2. it has been loaded from the storage

In case 2, the value must have been loaded with a tid <= current tid (the tid associated with a connection cache is strictly increasing).
Had we read `obj1` with that tid, we should have read the same value, i.e. `26`.
As we now read `obj1` as `25`, we can conclude `tid != current tid`, i.e. tid < current_tid`.
This would mean that the storage knows value `26` at time tid, but `25` at
time `current tid > tid` contradicting that the values are strictly increasing.

For case 1, the storage and its client cache are updated as part
of `tpc_finish` processing for transaction id (say) tid.
The updates would affect both `obj1` and `obj2`.
The failing check round would read data at `current_tid >= tid`
after `tpc_finish` processing is complete. It should therefore
see identical values for `obj1` and `obj2` -- both at the connection cache
as well as at the storage level.


I suggest to simplify the test (or add another one):
the simplified test would work with a single object (rather than 2).
this object is first loaded "normally" (e.g. potentially via
the connection cache), then invalidated and loaded again (e.g. bypassing
the (invalidated) value in the connection cache).
Differing values indicate a cache inconsistency.
The `serial`s associated with the differing values and the "current tid"
associated with the connection cache could provide valuable hints
for the inconsitency analysis.

navytux:
@d-maurer, thanks for feedback and suggestions. For now I hesitate to change the test to use only one object. In my previous experience exercising things with only one object interferes with stream of updates when we want to check for improper invalidations. But the suggestion to show details of obj1 and obj2 serials at the failure is good.

So I've amended the test by emitting those details and also zconn.at :

https://github.com/navytux/ZEO/blob/f2bc73ae23969a9e8c57bec914b78c55713915db/src/ZEO/tests/testZEO.py#L390-L616
https://github.com/navytux/ZEO/blob/f2bc73ae23969a9e8c57bec914b78c55713915db/src/ZEO/tests/testZEO.py#L2141-L2160

and, unfortunately, yesterday no matter how much I've tried, I could not reproduce the problem locally on my own computers.

But today, I could reproduce this problem again at least once on github actions:

https://github.com/navytux/ZEO/runs/7689288409?check_suite_focus=true

and it failed like this:

```
Failure in test check_ZZ_race_load_vs_external_invalidate (ZEO.tests.testZEO.BlobAdaptedFileStorageTests)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/runner/work/ZEO/ZEO/.tox/py27-zeo4/lib/python2.7/site-packages/ZODB/tests/util.py", line 400, in _
    return f(*argv, **kw)
  File "/home/runner/work/ZEO/ZEO/src/ZEO/tests/testZEO.py", line 616, in check_ZZ_race_load_vs_external_invalidate
    self.fail([_ for _ in failure if _])
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 410, in fail
    raise self.failureException(msg)
AssertionError: [u'T7: obj1.value (2)  !=  obj2.value (3)\nobj1._p_serial: 0x03ea127f67989ecc  obj2._p_serial: 0x03ea127f67a52e22  zconn.at: 0x03ea127f67a52e22']
```

At the moment of when the data corruption was detected zconn was  `@03ea127f67a52e22`. Obj2, whose state is more likely to be retrieved via `ZODB.Connection` cache, was loaded correctly, also with the same serial as zconn.at. But obj1, which we try to load afresh from the storage, was loaded with _older_ serial - `@03ea127f67989ecc` : 

```
In [1]: s1 = 0x03ea127f67989ecc

In [2]: s2 = 0x03ea127f67a52e22

In [3]: s1 < s2
Out[3]: True

In [4]: s2 - s1
Out[4]: 823126
```

which means that loading the object directly from ClientStorage got it wrong.

This load still uses ClientStorage cache. I will try to amend the test to get those details of the cache also into the failure. Currently, what comes to mind, is that maybe it is ClientCache invalidations in ZEO4 mode, is wrong - it works differently compared to when working with ZEO5: on ZEO4 the server calls `serialnos` to the client instead of returnnig committed serials from `tpc_vote`. More details about serialnos are in https://github.com/zopefoundation/ZEO/commit/020a9f8c671f .

navytux:
I could finally reproduce the problem locally. Here is e.g. the failure where obj1 turns out to be older than obj2:

```
T4: obj1.value (2)  !=  obj2.value (3)
obj1._p_serial: 0x03ea140e076bb799  obj2._p_serial: 0x03ea140e07dfef33  zconn.at: 0x03ea140e07dfef33
zcache.loadBefore(obj1, zconn.at)  ->  serial: 0x03ea140e076bb799  end_tid: None
zcache.loadBefore(obj2, zconn.at)  ->  serial: 0x03ea140e07dfef33  end_tid: None
zstor.loadBefore(obj1, zconn.at)   ->  serial: 0x03ea140e07dfef33  next_serial: None
zstor.loadBefore(obj2, zconn.at)   ->  serial: 0x03ea140e07dfef33  next_serial: None
```

The older value is loaded from ClientStorage cache. The storage itself is ok. It is, thus, in this case, ZEO cache invalidation logic which was wrong.

Digging more...

d-maurer:
Kirill Smelkov wrote at 2022-8-5 03:48 -0700:
> ...
>Currently, what comes to mind, is that maybe it is ClientCache invalidations in ZEO4 mode, is wrong - it works differently compared to when working with ZEO5: on ZEO4 the server calls `serialnos` to the client instead of returnnig committed serials from `tpc_vote`.

I like the hypothesis because it implies that only the
"ZEO5 client/ZEO 4 server" scenario has the problem.

However, I do not yet see how the race can occur:
neither `vote` nor `serialnos` perform invalidations.
Both update the associated `TransactionBuffer`
which is passed to the `tpc_finish` call
which in turn performs the invalidations.
The ZEO 4 server calls `client.serialnos` before
it returns from the vote call; this implies that
the `serialnos` message arrives at the client before
the `vote` result message and the `TransactionBuffer`
should have the same state at the end of the client's `tpc_vote`.

I see the following difference:
`serialnos` is an asynchronous call. As such, it has significantly
more problems to handle exceptions than the synchronous `vote`.
Looking at `ClientStorage.tpc_vote`, the case should be handled
by setting an exception on the `TransactionBuffer`.
But maybe, something is buggy in this regard.

I suggest to check for exceptions near the test failure point.
In the context of #195, activating logging helped a lot
to get a good overview of exceptions raised during tests.
`zope.testrunner` looks at `log.ini` in the current working directory
for the logging configuration.

Unfortunately, I do not know how to access the produced log file
on `github`.
Should you follow the approach on your local system,
you might require extensions in the newest `zope.testrunner`
and #195 in the case your system is MacOS (or Windows) based.
MacOS uses `spawn` to start new processes (such as the ZEO server)
and therefore, the process does not inherit the log configuration.
The extensions mentioned above work around this.

navytux:
@d-maurer, thanks for feedback. My system is Linux based and now, it seems, I can reliably reproduce the failure in several minutes. I will try to use your logging suggestion as needed.

Btw, serialnos` actually does invalidations:

https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/asyncio/client.py#L772-L788

but, suspecting the race at that place, I already tried to add `sleep(1)` in serialnos befre cache invalidations on the client, and also on the server before sending `serialnos` message. All that did not triggerred the issue, so, it seems, it is not related to suspected serialnos race.

d-maurer:
Kirill Smelkov wrote at 2022-8-5 09:54 -0700:
>I could finally reproduce the problem locally. Here is e.g. the failure where obj1 turns out to be older than obj2:
>
>```
>T4: obj1.value (2)  !=  obj2.value (3)
>obj1._p_serial: 0x03ea140e076bb799  obj2._p_serial: 0x03ea140e07dfef33  zconn.at: 0x03ea140e07dfef33
>zcache.loadBefore(obj1, zconn.at)  ->  serial: 0x03ea140e076bb799  end_tid: None
>zcache.loadBefore(obj2, zconn.at)  ->  serial: 0x03ea140e07dfef33  end_tid: None

This is very strange:
All invalidations for a transaction should get processed together.
As the test modifies `obj1` and `obj2` always together in the same
transaction, the corresponding invalidations should arrive in
the same invalidation message or be performed in the same `tpc_finish`
call.

The invalidations for the individual objects in an invalidation
message or a `tpc_finish` call are not performed atomically.
Therefore, it is possible in principle that one object
is already invalidated but not yet the other one.
However, the invalidations should apply to a state in the future
(more recent than `zconn.at`) while the application accesses
the state at `zconn.at`. All invalidations corresponding to
`zconn.at` should have been processed completely - i.e.
either for both `obj1` and `obj2` or for none of them.

navytux:
Yes, it is indeed strange, but it already happened on my system ~ 5 times. There is one suspicious place, that @jmuchemb was talking about in https://github.com/zopefoundation/ZEO/issues/155#issuecomment-577314441 about https://github.com/zopefoundation/ZEO/commit/062cbecc273b58ebbebd26eec1e8abae89795d0d : 

https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/ClientStorage.py#L857-L868

because that codepath invalidates only one object on `ConflictError`, but I've added the following prints to see whether that codepath gets activated or not:

```diff
         except POSException.ConflictError as err:
+            print('\n\n\n')
+            print('vote: ConflictError', err)
             oid = getattr(err, 'oid', None)
             if oid is not None:
+                print('vote: ConflictError -> invalidate oid')
                 # This is a band-aid to help recover from a situation
                 # that shouldn't happen.  A Client somehow misses some
                 # invalidations and has out of date data in its
@@ -865,14 +872,19 @@ def tpc_vote(self, txn):
                 # (unresolved) conflict error, we assume that the
                 # cache entry is bad and invalidate it.
                 self._cache.invalidate(oid, None)
+            print('\n\n\n')
             raise
```

and I do not see it being executed.

For the reference: I've made sure that regular print output gets emitted from that file during the test run ok. (I run the test via  `zope-testrunner -fvvvx --test-path=src -t ZZ_race_load_vs_external_invalidate`).

d-maurer:
Kirill Smelkov wrote at 2022-8-5 10:18 -0700:
***@***.***, thanks for feedback. My system is Linux based and now, it seems, I can reliably reproduce the failure in several minutes. I will try to use your logging suggestion as needed.
>
>Btw, serialnos` actually does invalidations:
>
>https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/asyncio/client.py#L772-L788

I find this behavior doubtful:
`serialnos` is called as part of `vote`; correspondingly, the values for the
respective objects have not yet been stored in the cache
(this would happen in `tpc_finish` if no exception occured).
I do not see a reason to perform a special case invalidation
in `serialnos` (beside the normal invalidation processing).

On the other hand, it should not explain the observed corruptions
even though it is a difference between ZEO4 and ZEO5 server behavior
(with a ZEO 5 server, there would be no invalidations during `tpc_vote`).

d-maurer:
Kirill Smelkov wrote at 2022-8-5 10:45 -0700:
>Yes, it is indeed strange, but it already happened on my system ~ 5 times. There is one suspicious place, that @jmuchemb was talking about in https://github.com/zopefoundation/ZEO/issues/155#issuecomment-577314441 about https://github.com/zopefoundation/ZEO/commit/062cbecc273b58ebbebd26eec1e8abae89795d0d :
>
>https://github.com/zopefoundation/ZEO/blob/fc0729b3cc754bda02c7f54319260b5527dd42a3/src/ZEO/ClientStorage.py#L857-L868
>
>because that codepath invalidates only one object on `ConflictError`, but I've added the following prints to see whether that codepath gets activated or not:
>
>```diff
>         except POSException.ConflictError as err:
>+            print('\n\n\n')
>+            print('vote: ConflictError', err)
>             oid = getattr(err, 'oid', None)
>             if oid is not None:
>+                print('vote: ConflictError -> invalidate oid')
>                 # This is a band-aid to help recover from a situation
>                 # that shouldn't happen.  A Client somehow misses some
>                 # invalidations and has out of date data in its
>@@ -865,14 +872,19 @@ def tpc_vote(self, txn):
>                 # (unresolved) conflict error, we assume that the
>                 # cache entry is bad and invalidate it.
>                 self._cache.invalidate(oid, None)
>+            print('\n\n\n')
>             raise
>```
>
>and I do not see it being executed.

As the comment above suggests, this code is there only as
a desperate (precautionary) measure to counter a previous
error (lost invalidation) which should be impossible in the first place.
But it is executed whenever `vote` raises a `ConflictError`
with associated `oid` information, i.e. much more often.

As a side note: it is possible that the code above is not
executed at all for the scenario involving a ZEO 4 server.
When I understood the ZEO 4 server code correctly, then
write failures are reported as part of the `serialnos` call,
not necessarily via an exception from `vote`.
As you have pointed out earlier, the `serailnos` implementation
performs similar and even much wider invalidations.


The code also proves wrong one of my previous statements:
"In the scenario "ZEO 5 client/ZEO 5 server", `ClientStorage`
does not do invalidations during `tpc_vote`".

This behavior is a conceptual flaw: all cache modifications
have been moved into the IO thread -- apart from the
invalidation above.
However, cache modifications are lock protected;
the invalidation above therefore should not introduce a problem
even though it is executed in the wrong thread.

navytux:
Ok, please see https://github.com/zopefoundation/ZEO/issues/209 for the analysis of what it turned out to be.

PR #210, opened 2022-09-01
Run tests at all levels

navytux:
Since https://github.com/zopefoundation/ZODB/commit/a94d63c552d6 (part
of https://github.com/zopefoundation/ZODB/pull/368) some ZODB tests are
marked as being long to help skipping them easily during development via
running tests only at level 1.

But by default we want to continue to run all tests.

-> Adjust tox to run tests at all levels explicitly, since
zope-testrunner's default is to run tests at level 1 only.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,9 +28,9 @@ setenv =
     zeo4: ZEO4_SERVER=1
 commands =
     # Run unit tests first.
-    zope-testrunner -u --test-path=src --all {posargs:-vc}
+    zope-testrunner -u --test-path=src -a 1000 {posargs:-vc}
     # Only run functional tests if unit tests pass.
-    zope-testrunner -f --test-path=src --all {posargs:-vc}
+    zope-testrunner -f --test-path=src -a 1000 {posargs:-vc}
 extras =
     test
 
--- a/tox.ini
+++ b/tox.ini
@@ -28,9 +28,9 @@ setenv =
     zeo4: ZEO4_SERVER=1
 commands =
     # Run unit tests first.
-    zope-testrunner -u --test-path=src {posargs:-vc}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -28,9 +28,9 @@ setenv =
     zeo4: ZEO4_SERVER=1
 commands =
     # Run unit tests first.
-    zope-testrunner -u --test-path=src {posargs:-vc}
+    zope-testrunner -u --test-path=src --all {posargs:-vc}
```
d-maurer, line 31: Apparently, `--all` takes precedence over `-a`. This would imply that the expensive tests cannot be avoided for `tox` runs.

I see two potential workarounds:
1. use `-a 0` instead of `-all`; because the argument parser allows multiple `-a` options and takes the last value, the `-a 0` from `tox.ini` could be overridden with a `-a 1`
2. move the `-all` into the `posarg` default; this assumes that in `{`*var*`:`*xxx*`}`, the *xxx* specifies the default to be used if *var* is empty.

navytux: @d-maurer thanks for feedback and good point. I believe the first option is better, because on the second on, if _any_ positional argument is used, then the default settings for level will be deactivated. The only difference is that I used `-a 1000` instead of `-a 0`, because contrary to what zope.testrunner documentation says, `-a 0` actually _disables_ tests at all levels. Please see https://github.com/zopefoundation/ZEO/pull/210/commits/06f8e1461c05b3664512eac533947fd38d266946 for details. Hope it is ok.

P.S. after settling here we should adjust .meta and tox.ini in ZODB.git correspondingly.

d-maurer (approved):

navytux:
@d-maurer, thanks for the approval.

Issue #209, opened 2022-08-15
Multiple ZEO data corruptions due to concurrency bugs


navytux:
Hello up there. In https://github.com/zopefoundation/ZEO/pull/207 it was [noted](https://github.com/zopefoundation/ZEO/pull/207#issuecomment-1202260612) that current ZEO5 master fails sometimes in `check_race_load_vs_external_invalidate` wrt ZEO4 server. I digged in to find out why and discovered multiple concurrency bugs - in both ZEO4 and ZEO5 - that lead to data corruptions. Please find details about those bugs below:

## Bug1: ZEO5.client - ZEO4.server : race in between load / invalidate

The first bug is that ZEO5.client - contrary to ZEO4.client - does not account for simultaneous invalidations when working wrt ZEO4.server.
It shows as e.g.

```console
(z-dev) kirr@deca:~/src/wendelin/z/ZEO5$ ZEO4_SERVER=1 zope-testrunner -fvvvx --test-path=src -t check_race_load_vs_external
...

AssertionError: T1: obj1 (24)  !=  obj2 (23)
obj1._p_serial: 0x03ea4b6fb05b52cc  obj2._p_serial: 0x03ea4b6faf253855
zconn_at: 0x03ea4b6fb05b52cc  # approximated as max(serials)
zstor.loadBefore(obj1, @zconn.at)       ->  serial: 0x03ea4b6fb05b52cc  next_serial: None
zstor.loadBefore(obj2, @zconn.at)       ->  serial: 0x03ea4b6faf253855  next_serial: 0x03ea4b6fb07def66
zstor._cache.clear()
zstor.loadBefore(obj1, @zconn.at)       ->  serial: 0x03ea4b6fb05b52cc  next_serial: 0x03ea4b6fb07def66
zstor.loadBefore(obj2, @zconn.at)       ->  serial: 0x03ea4b6fb05b52cc  next_serial: 0x03ea4b6fb07def66
```

indicating that obj2 was provided to user from the cache that erroneously got stale.

With [IO trace](https://lab.nexedi.com/kirr/ZEO/commit/f184a1d9) showing message exchange in between client and the server, this looks as:

```
# loadBefore issued
tx (('\x00\x00\x00\x00\x00\x00\x00\x02', '\x03\xeaKo\xaf%8V'), False, 'loadBefore', ('\x00\x00\x00\x00\x00\x00\x00\x02', '\x03\xeaKo\xaf%8V'))

# received invalidateTransaction
[...]

PR #211, opened 2022-09-01
Remove multi-threaded server mode

navytux:
Since 2016 this mode was opt-in for a long time with the default being
single-threaded asyncio-based server implementation. We believe nobody
actually used multi-threaded ZEO5 server mode for real at all.

Multi-threaded server mode was deprecated this spring
(https://github.com/zopefoundation/ZEO/pull/190), and later found to
have concurrency bug that lead to data corruption
(https://github.com/zopefoundation/ZEO/issues/209).

Given deprecation status, instead of fixing those bugs let's finally
remove the multi-threaded server mode.


dataflake (approved):

navytux:

Issue #214, opened 2022-09-30
zeopack exit value should not be 0 when socket is not found


sblondon:
ZEO 5.3 is installed in a virtualenv, from pypi.org. We use `zeopack` to pack the database with this command, inside a shell script:

```
/path/to/zeopack -u /path/to/zeo.sock:1
```

If `/path/to/zeo.sock` does not exist (because ZEO server does not run), the output shows the problem:

```
Couldn't connect to: ('/path/to/zeo.sock', '1')
```

It's good but the exit value is 0. I would expect another value like 1 (which is the value returned by the [`error()` function](https://github.com/zopefoundation/ZEO/blob/83730901e5d7bcb0916d5391f5907e5f9d3a32e7/src/ZEO/scripts/zeopack.py#L80)).
My usecase is to get the exit value to modify the behaviour of the shell script.

The error occurs in [a for loop](https://github.com/zopefoundation/ZEO/blob/83730901e5d7bcb0916d5391f5907e5f9d3a32e7/src/ZEO/scripts/zeopack.py#L148) trying to get a connexion on a `ZEO.ClientStorage.ClientStorage()`. I guess the user can provide several servers so there is an attempt on each server until one connexion succeed.
I think about two ways to fix it:
- a. If several servers can be configured, there can be a case at the end of the function if all servers failed
- b. If only one server can be configured, the [`continue`](https://github.com/zopefoundation/ZEO/blob/83730901e5d7bcb0916d5391f5907e5f9d3a32e7/src/ZEO/scripts/zeopack.py#L160) can be replaced by an `error()` call.

What do you think about such solutions?
I can do a PR to implement it (or another way to fix it if you have a better strategy).

For the record, I found this behaviour with cPython3.9 and ZEO-5.3.0 but this part of the code matches the code in the repository in `master` branch.

PR #215, opened 2022-10-03
exit value is 1 if all connection to servers fail

sblondon:
closes: #214

This PR changes the `zeopack` exit value to 1 when all connections to server are a failure. Currently, `zeopack` outputs `Couldn't connect to: xxx` and exits with 0. I think it should be an error exit value than a succes. 

The usecase is described in #214. This PR implements  the solution **a** (as defined in issue #214) because the [`usage`](https://github.com/zopefoundation/ZEO/blob/83730901e5d7bcb0916d5391f5907e5f9d3a32e7/src/ZEO/scripts/zeopack.py#L12) constant documents `zeopack` as `Pack one or more storages hosted by ZEO servers`.

### DIFF
--- a/src/ZEO/scripts/zeopack.py
+++ b/src/ZEO/scripts/zeopack.py
@@ -169,6 +169,7 @@ def _main(args=None, prog=None):
     if not server_found:
         error("Cannot connect to any of the configured ZEO servers.")
 
+
 def main(*args):
     root_logger = logging.getLogger()
     old_level = root_logger.getEffectiveLevel()
--- a/src/ZEO/scripts/zeopack.py
+++ b/src/ZEO/scripts/zeopack.py
@@ -145,12 +145,14 @@ def _main(args=None, prog=None):
     if not servers:
         error("No servers specified.")
 
+    server_found = False
     for addr, name in servers:
         try:
             cs = ZEO.ClientStorage.ClientStorage(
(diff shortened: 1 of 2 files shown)


src/ZEO/scripts/zeopack.py
```diff
@@ -164,6 +166,8 @@ def error(message):
             traceback.print_exception(*(sys.exc_info()+(99, sys.stderr)))
             error("Error packing storage %s in %r" % (name, addr))
 
+    if not server_found:
+        error("No servers found.")
```
dataflake, line 170:
```suggestion
        error("Cannot connect to any of the configured ZEO servers.")
```

dataflake (changes requested):
Please change the error message as suggested to make the problem more obvious.

dataflake:
Don't forget to fix the failing tests. There were some looking for the specific output generated by the old code. They are all failing now.

sblondon:
@dataflake thank for the review. The tests are fixed now.

dataflake (approved):

icemac:
Should there be a message in the change log or is this change merely internally?

dataflake:
I added an entry in https://github.com/zopefoundation/ZEO/commit/a6072ecd2aceb004187fdd718b8f1c3c64ffa0e6

PR #200, opened 2022-05-18
Drop Python2 support  (WIP: do not merge before #208 or #195)

navytux:
--------
kirr:

ZEO6 is going to support Python3 - see discussion around https://github.com/zopefoundation/ZEO/pull/195#issuecomment-1093837395 .

This patch was extracted from https://github.com/zopefoundation/ZEO/pull/195 .

PR #218, opened 2023-01-08
`ZEO.asyncio`: switch to `async/await` style

navytux:
Second part of Dieter's work from https://github.com/zopefoundation/ZEO/pull/195 :

- Switch to using `async/await` directly instead of `@coroutine/yield`
- Drop Python 2 and 3.5 support


navytux:
It is a question whether we want to merge this - from https://github.com/zopefoundation/ZEO/pull/195#issuecomment-1374901652:

2. the second part drops Python2 support and switches from async/await-style
   coroutines being implemented with `@coroutine/yield` syntax to those
   coroutines implemented with `async/await` syntax directly. Due to the way
   how part 1 is organized this change is small. It is represented by
   https://github.com/zopefoundation/ZEO/commit/7f574ecd patch and further
   changes that drop py2 support otherwise in `knext+py3` branch and
   https://github.com/zopefoundation/ZEO/pull/218 pull-request.

   In https://github.com/zopefoundation/ZEO/pull/200#issuecomment-1202136123 you
   said that _"dropping Python 2 support gives no gain by itself. Only if
   gainfull things (I think of #195) require dropping Python 2 support, I would
   merge #200."_ So from this point of view, if maybe we are ok to keep our
   `async`/`await`-style coroutines being implemented with similar
   `@coroutine`/`yield` syntax, there is no real need to merge this and to drop
   py2 support.

d-maurer:
Kirill Smelkov wrote at 2023-1-8 10:50 -0800:
> ...
>   In https://github.com/zopefoundation/ZEO/pull/200#issuecomment-1202136123 you
>   said that _"dropping Python 2 support gives no gain by itself. Only if
>   gainfull things (I think of #195) require dropping Python 2 support, I would
>   merge #200."_ So from this point of view, if maybe we are ok to keep our
>   `async`/`await`-style coroutines being implemented with similar
>   ***@***.***`/`yield` syntax, there is no real need to merge this and to drop
>   py2 support.

I got the feeling that all `zopefoundation` packages will
drop Python 2 support the next time, they are reconfigured with
the `meta` tools.

Thus, after you have merged the Py2/Py3 changes, you should
likely make a 5.x branch.

navytux:
Thanks, I see.

dataflake:
We are in the slow process of dropping support for all Python versions up to and including Python 3.6. Many dependencies have already done that and the pain of supporting those old Python versions is now higher than any gain.

navytux:
I see, Jens. Thanks for clarifying. So from this point of view it indeed makes sense to want to merge this PR.

dataflake:
IMHO you should also remove Python 3.6 support before merging.

navytux:
> IMHO you should also remove Python 3.6 support before merging.

@dataflake, would you please clarify about why? Currently supporting py3.6 does not add big extra cost.

dataflake:
It is EOF and we're removing support on all packages we touch in the zopefoundatiom organization.

navytux:
Thanks, Jens. I've dropped support for py3.6 in https://github.com/zopefoundation/ZEO/pull/218/commits/7ecc6a39.

navytux:
( https://github.com/zopefoundation/ZEO/actions/runs/3949781601/jobs/6761437520 says all tests passed but the status on PR is still "waiting for py27"; probably it is something in project settings )

dataflake:
I changed the project settings, please proceed

navytux:
(all tests passed ok now; Jens thanks once again)

src/ZEO/asyncio/_futures.pyx
```diff
@@ -303,13 +271,6 @@ cdef class CoroutineExecutor:
...
-            elif PY_MAJOR_VERSION < 3 and isinstance(result, asyncio.Future):
-                await_next = result  # trollius predates ._asyncio_future_blocking
-
-            # `yield coro` - handle as if it was `yield from coro`
-            elif _iscoroutine(result):
-                # NOTE - always AsyncTask even if we are originally under ConcurrentTask
-                await_next = AsyncTask(result, self.task.get_loop())
 
             else:
                 # object with __await__ - e.g. @cython.iterable_coroutine used by uvloop
```
d-maurer, line 276: Why do we need this special `_await__` logic? It is not present in #195 and nevertheless all `uvloop` tests passed.

src/ZEO/asyncio/marshal.py
```diff
@@ -21,10 +21,9 @@
 
 import logging
 
-from .._compat import Unpickler, Pickler, BytesIO, PY3
+from .._compat import Unpickler, Pickler, BytesIO
```
d-maurer, line 24: Why do we import `BytesIO` from `_compat`? The `_compat` definition comes from `ZODB._compat` which uses `io.BytesIO` for PY3 and `cStringIO.StringIO` for PY2.

d-maurer (commented):
Very thorough removal of Py2 code!

The reason why I do not yet approve is the special `__await__` code in the `CoroutineExecutor` implementations. I think, we do not need it.

d-maurer:
> The reason why I do not yet approve is the special `__await__` code in the `CoroutineExecutor` implementations. I think, we do not need it.

There is one important difference between your implementation and that of #195: #195 uses `loop.create_task` to create tasks during connection establishment and the special ZEO tasks only after the connection is fully established; your implementation uses the ZEO tasks throughout. This might explain why your `CoroutineExecutor` faces loop (i.e. `uvloop`) specific implementation details, not seen by #195.

In my view, this would suggest that `loop.create_task` is more adequate at least for the initial connection phase which might be tightly coupled with the loop implementation. We know that calling callbacks immediately is no problem for the ZEO server calls but the connection establishment might not anticipate this.

.meta.toml
```diff
@@ -27,10 +27,20 @@ testenv-deps = [
     "uvloop: uvloop",
     ]
 testenv-setenv = [
-    "!py27-!pypy: PYTHONWARNINGS=ignore::ResourceWarning",
+    "PYTHONWARNINGS=ignore::ResourceWarning",
```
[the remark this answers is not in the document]
dataflake: In 927fcdd35cf048616bb8554ca03b5005f8d5e7e3 I have removed that setting. `ResourceWarnings` are no longer ignored. It was introduced early on when ZEO `master` tests spewed hundreds of them.

src/ZEO/asyncio/_futures.pyx
navytux: Dieter, thanks for noticing. I simply overlooked to remove this in 7f574ecd. I removed that code in https://github.com/zopefoundation/ZEO/pull/218/commits/32091b5f. Thanks, once again, for pointing this out.

src/ZEO/asyncio/marshal.py
navytux: Simply because there was not BytesIO import fixes in #195 and I was doing things only based on that. But you are right - I removed py2 compatibility handling for BytesIO and StringIO throughout in https://github.com/zopefoundation/ZEO/pull/218/commits/1f16336c.

dataflake:
@navytux I just added a very large commit with more Python 2/3 cleanups, make sure to pull first

navytux:
Dieter, thanks for feedback and kind words.

> Very thorough removal of Py2 code!

Thanks. This is the power of doing one thing at a time in every step and doing it in
full. And it is even relatively easy to do with e.g. `git grep -Ovi` and similar ways.


> The reason why I do not yet approve is the special __await__ code in the CoroutineExecutor implementations. I think, we do not need it.

You are right. As explained in https://github.com/zopefoundation/ZEO/pull/218#discussion_r1081166275 I removed this in https://github.com/zopefoundation/ZEO/commit/32091b5f.

For py2 it was the only way to make things working by using our own futures
and `yield from` emulating code in the CoroutineExecutor. For py3 it might work
both ways - via native futures and via our futures. But since our futures way
already works, and we removed the code, that was worrying you, from
CoroutineExecutor, I believe there is no downside in using our infrastructure to
execute all coroutines.

I think after we removed that extra py2 supporting code from CoroutineExecutor,
the difference you point out should not be a blocker. I believe it is better to
use our own Futures and Tasks throughout uniformly - so that they are all
properly tested and we are sure that it works all ok. I think it would create a
bit of potential confusion if in some cases we use native asyncio futures,
while in other cases we use our own futures. It would then be necessary to
keep remembering which is used where and care about that.


> @navytux I just added a very large commit with more Python 2/3 cleanups, make sure to pull first

Jens, thanks for notifying. I went briefly through that large update and
noticed something strange in changes to `src/ZEO/hash.py`:

https://github.com/zopefoundation/ZEO/pull/218/commits/bac6d86f3dbac39beaaa597f7af5a98e45618fef#diff-ef327459a306a2207e1a6d8ed5d2959a40c7ab9de2d15a1139e861df3a2f6293

if that code is not used at all I believe we should simply remove the whole
file instead of deactivating the code inside.

Do you agree?


Kirill

dataflake:
> Jens, thanks for notifying. I went briefly through that large update and
noticed something strange in changes to src/ZEO/hash.py:
>
> https://github.com/zopefoundation/ZEO/commit/bac6d86f3dbac39beaaa597f7af5a98e45618fef#diff-ef327459a306a2207e1a6d8ed5d2959a40c7ab9de2d15a1139e861df3a2f6293
> 
> if that code is not used at all I believe we should simply remove the whole
file instead of deactivating the code inside.

Of course that file can go, I had commented it out to see if it makes a difference and it obviously didn't. That was an oversight.

navytux:
Thanks, Jens. I removed it in https://github.com/zopefoundation/ZEO/pull/218/commits/b4507e55.

navytux:
Btw, Dieter, any idea about https://github.com/zopefoundation/ZEO/actions/runs/3958283359/jobs/6779667424 ?

d-maurer:
> I think after we removed that extra py2 supporting code from CoroutineExecutor,
> the difference you point out should not be a blocker. I believe it is better to
> use our own Futures and Tasks throughout uniformly - so that they are all
> properly tested and we are sure that it works all ok.

Our tasks not only execute callbacks immediately, they ignore also the task context (used e.g. for context variables). We know that our coroutines make no use of the task context; however, a priori, we know nothing about the "create connection coroutine". In my opinion, it would therefore be safer to use `loop.create_task` at least for the "create connection coroutine" as this guarantees a task which provides all of the standard `asyncio.Task` functionality (including full context support).

navytux:
Dieter, thanks for feedback.

Ok, let it be this way. I've changed `._connecting` to be created via
`loop.create_task` as you suggest: https://github.com/zopefoundation/ZEO/pull/218/commits/63429391.

Kirill

d-maurer:
> Btw, Dieter, any idea about https://github.com/zopefoundation/ZEO/actions/runs/3958283359/jobs/6779667424 ?

This is a race condition caused by a buggy test:
the test sets up a `ClientStorage` for storage `2` -- which the server does not have. It then closes the storage and checks the log records. The test succeeds when the server response about the unknown storage has not yet been processed (this is typically the case); otherwise, the log records associated with this processing let the test fail.

The log record examination exists primarily to help in the analysis of earlier problems with the closing logic: in some cases, the `close` deadlocked. To work around this, I introduced a timeout into the closing and the generation of a log message when it terminates the closing process. The log record examination should detect  a corresponding log record but actually does more. I suggest to replace it with
```pycon
    >>> def check_log_records(records):
    ...     records = [r.message for r in records]
    ...     return any("bug" in r for r in records) and records else []
```

d-maurer (approved):

navytux:
Dieter, thanks for the approval and for explaining what is going on with `check_log_records`.

>  I suggest to replace it with
>
> ```
> >>> def check_log_records(records):
>     ...     records = [r.message for r in records]
>     ...     return any("bug" in r for r in records) and records else []
> ```

I agree. Would you please do this, probably with also putting a comment regarding the race you describe? I mean this is your test and your code, and it was you to analyze what is going on here. It looks logical for you to make the fixing commit.

----

@dataflake, are you also ok with merging this PR after we do the above?

d-maurer:
Kirill Smelkov wrote at 2023-1-20 00:52 -0800:
> ...
>I agree. Would you please do this ...

Ok -- when I have finished my current project.

navytux:
Thanks.

navytux:
Dieter, thanks for the fixups. Do I understand it right that this PR should be ok to be merged now?

navytux:
> @dataflake, are you also ok with merging this PR after we do the above?

Jens, we seem to be done. Do you agree with merging this?

d-maurer (approved):

dataflake:
Yes you can go ahead and merge

dataflake (approved):

navytux:
Dieter, Jens, thanks for feedback and approvals. I've merged this via https://github.com/zopefoundation/ZEO/commit/59de2de3.

Using the occassion I have a question regarding 5.x branch: now since hereby PR is merged, I want to pick some fixes and improvements from master to 5.x . For example I would like to pick e.g. d08ac83b, removal of `ZEO/zhash.py`, 3ca79a69, changing deprecated threading API to up-to-date ones (e.g. `setDaemon(True)` -> `.daemon = True`), etc. Do I need to open PR for such backports, or is it ok if I do it in 5.x directly by myself?

dataflake:
I think it's fine if you do this by yourself.

navytux:
Thanks, Jens. I picked up the following patches to 5.x for now:

- 3a38db4821a72279e95b145ee2ec82de874d4b7a
- ed63f52f15b92cb6ff6501c8e5187fe46ad4a040
- 40ee0cffd1fd9366ae204532b32569b3842aef40
- abd1c655c71e656d3b5c776a8a817967f2e76493
- dce15bea0b8e20de91ed8a1bba7ee6cb1ecfb335
- 421996d4a3c4e61680e988cb4342927fae4c0f38
- b3d13dd1f36dd093d36ddf8e3e061c28f3d64279
- 5293b3c3222ec0e357c62e3e9a4f3f5ae36c2d97

I did not picked change of `%` to `.format` because sometimes you changed to `.format` and sometimes to using f-strings. And even though `.format` is supported on both py2 and py3, f-strings are supported only on py3. I wanted to keep the diff in between 5.x and master minimal for easier tracking, but regarding strings I'm not sure it makes sense to pick up only partial change to `.format`.

Hope my backports are ok.

dataflake:
You do not need to verify every detail with me. I don't have much interest in the ZEO `5.x` branch. My goal is to move the main "consumer" of ZEO, Zope, to ZEO 6 as soon as the first release is there.

navytux:
I see. Thanks for feedback.

PR #219, opened 2023-01-08
[for discussion] asyncio.client: move cache access out of `ClientIO` into `Protocol`

navytux:
to ensure that we do not store the same information twice into the cache; this may or may not be important

--------
kirr:

	for discussion

### DIFF
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -328,18 +328,32 @@ class Protocol(base.ZEOBaseProtocol):
         future = self.futures.get(message_id)
         if future is None:
             future = Future(loop=self.loop)
-            self.call_sync('loadBefore', message_id, message_id, future)
-
-            @future.add_done_callback
-            def _(future):
-                try:
+            # Check whether the cache contains the information.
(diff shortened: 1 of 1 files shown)


navytux:
See _"a) moving cache access out of ClientIO into Protocol ..."_ in https://github.com/zopefoundation/ZEO/pull/195#issuecomment-1374901652 .

d-maurer:
#195 does not try to move the cache access out of `ClientIO` to the protocol but have it in the protocol as well.

The cache access in `ClientIO` is there to be able to deliver cached data even when the server connection is down (i.e. we do not have a protocol instance). In real life, this is likely not necessary because `ClientStorage` has already tried a cache access and will only use lower level calls when this has failed. However, there is a test which verifies the behavior.

The cache access in `ClientIO` usually makes the one in `Protocol.load_before` redundant, unless the client has not been "operational" (i.e. did not have an associated protocol). In the latter case, the `ClientIO` cache access and the one in `load_before` do not happen in the same coroutine step and therefore, the cache might have been modified in the meantime.

I do not know what happens when information is stored in the cache which is already there. The original code tried hard to avoid this - maybe, it is important.

PR #223, opened 2023-01-23
Raise the open files limit if needed during test runs

### LINKED ISSUE #222: Race test `check_race_external_invalidate_vs_disconnect` fails on macOS
The race test `check_race_external_invalidate_vs_disconnect`, which is defined in the ZODB package, fails reliably on macOS regardless of the backing storage used by the test. The failure is `OSError: [Errno 24] Too many open files`.

The test definition in the module `ZODB/tests/racetests.py` sets a hardcoded number of 64 workers. That is too much for macOS. Manual testing has shown that dialing it down to 43 workers will make the tests run successfully.

The test itself is "re-used" by ZEO. Running the ZODB tests by themselves doesn't show the issue, 64 workers is fine in that case. I wonder if there's some elegant way of moving that worker limit setting somewhere else so it can be changed temporarily from within the ZEO tests if e.g. `sys.platform == 'darwin'`.

dataflake:
Fixes #222 

macOS uses a very low soft limit for open files (256), and that's not enough for the most brutal race tests like `check_race_external_invalidate_vs_disconnect`.

### DIFF
--- a/src/ZEO/tests/testZEO.py
+++ b/src/ZEO/tests/testZEO.py
@@ -13,6 +13,7 @@
 ##############################################################################
 """Test suite for ZEO based on ZODB.tests."""
 import multiprocessing
+import resource
 
 from ZEO.ClientStorage import ClientStorage
 from ZEO.Exceptions import ClientDisconnected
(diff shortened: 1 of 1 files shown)


navytux (approved):
Thanks, Jens. This looks reasonable to me.

PR #224, opened 2023-01-26
Optimize via direct socket access

d-maurer:
@navytux First trial to implement your idea to optimize via direct socket access.

The SSL tests have still severe problems (not completely unexpected - as they might use the socket a bit differently). Some other tests fail as well.

### DIFF
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -207,17 +207,18 @@ class Protocol(base.ZEOBaseProtocol):
 
     def connection_made(self, transport):
         logger.debug('connection_made %s', self)
-        write_lock = threading.Lock()
-        ori_write = transport.write
-        ori_write_ready = transport._write_ready
-        def write(data):
-            with write_lock:
-                ori_write(data)
-        def _write_ready():
-            with write_lock():
-                ori_write_ready()
-        transport.write = write
-        transport._write_ready = _write_ready
+        if self.client.direct_socket_access:
+            write_lock = threading.Lock()
+            ori_write = transport.write
+            ori_write_ready = transport._write_ready
+            def write(data):
+                with write_lock:
+                    ori_write(data)
+            def _write_ready():
+                with write_lock():
+                    ori_write_ready()
+            transport.write = write
+            transport._write_ready = _write_ready
         super().connection_made(transport)
         self.heartbeat(write=False)
 
@@ -232,9 +233,13 @@ class Protocol(base.ZEOBaseProtocol):
                     f.set_exception(ClientDisconnected(exc or "Closed"))
         else:
             client = self.client
+            # We must first inform the client about the disconnection
+            # before the pending futures are finalized in ``close``.
+            # This ensures that subsequent API calls see the disconnected
+            # state and can wait for a reconnection
+            client.disconnected(self)
             # will set ``self.client = None``
             self.close(exc or "Connection lost")
-            client.disconnected(self)
 
     verify_task = None
 
@@ -541,6 +546,9 @@ class ClientIO:
         self.cache = cache
         self.register_lock = asyncio.Lock()
         self.closing_protocol_futures = set()  # closing futures
+        # check whether we want to access the socket directly
+        #   for the moment, not for SSL
+        self.direct_socket_access = ssl is None
         self.disconnected(None)
 
     def new_addrs(self, addrs):
@@ -750,8 +758,13 @@ class ClientIO:
                 # to interact with the server (deadlock or
                 # ``ClientDisconnected`` would result).
                 self.client.notify_connected(self, info)
-                self.connected.set_result(None)  # signal full readyness
+                # Order is important: as soon as we set the result
+                # of ``connected``, a waiter for connectedness
+                # can proceed and expect ``operational == True``
+                # We are at this time fully operational even if
+                # the ``connected`` result is not yet set
                 self.operational = True
+                self.connected.set_result(None)  # signal full readyness
 
     def get_peername(self):
         return self.protocol.get_peername()
@@ -954,7 +967,8 @@ class ClientRunner:
         self.call_sync_fco = client.call_sync_fco
         self.load_before_fco = client.load_before_fco
         run_threadsafe = run_coroutine_threadsafe
-        run_fast = run_coroutine_fast
+        run_fast = client.direct_socket_access and run_coroutine_fast \
+                   or run_threadsafe
 
         def io_call(coro, wait=True, fast=True):
             """run coroutine *coro* in the IO thread.
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -879,22 +879,26 @@ class ClientIO:
         # -- see comment in ``call_sync_fco``
         return await self.protocol.load_before(oid, tid)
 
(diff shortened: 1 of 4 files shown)


d-maurer:
It seems too difficult to implement the idea for SSL connections:
Ior a "normal" transport, the logic between the application (`ZEO` in our case) and the socket is fairly simple (essentially a buffer with flow control support).  For SSL, it is much more complex, partly implemented by an `SSLPipe` and partly by the `SSLProtocol` itself. In particular, writing can require a `resume_reading`. This whole logic currently assumes to be run only by the IO thread. Breaking this assumption brings a huge risk for breakage.

I will check whether a switch is possible: optimize for non-SSL connections; do not optimize for SSL connections.

navytux:
Dieter, thanks for starting this. I will try to look after I finish my current project.

Though one thing I noted offhand is that for SSL, after initial TLS handshake is complete, a write might require read for re-negotiation:

https://www.openssl.org/docs/man1.1.1/man3/SSL_write_ex.

    ... As at any time a re-negotiation is possible, a call to a write function can also cause read operations! ...
    
This way maybe if we disable re-negotiation (`OP_NO_RENEGOTIATION`) we can count on write never triggering a need for read. After all TLS Record Protocol is very simple and in the protocol itself there is no need to want to read if one side does write at application level.

Kirill

d-maurer:
Kirill Smelkov wrote at 2023-1-26 13:27 -0800:
> ...
>This way maybe if we disable re-negotiation (`OP_NO_RENEGOTIATION`) we can count on write never triggering a need for read. After all TLS Record Protocol is very simple and in the protocol itself there is no need to want to read if one side does write at application level.

You are the one who stresses safety and wants to avoid potential problems --
especially ones which occur only under rare special conditions
and are really hard to track down.

I see complex code -- designed under the assumption that
it is run in a single thread. We try to run it from different
threads. This poses a risk even if we can exclude that the main code
is run concurrently. Some data structures can be used in common by our
main code and other activity, not easily controllable.
Violating the design assumption "run in a single thread" may
bring those data structures into an inconsitent state.


Even the simple case (no SSL connection) has a potential
risk: it contains `loop._add_writer` which modifies a
data structure used for both reading and writing.
Our lock prevents `loop._add_write` to run concurrently for
the same socket, but in principle concurrent operations for reading
could modify the data structures used in common.
I think that this will not pose a serious problem -- but I am not sure.


The SSL case is much more complex.
It is far harder to determine which structures are used for
both reading and writing and whether concurrent modifications
are safe.
I agree with you that in principle it should be possible
to use the socket from application threads, too.
But, I am far less certain that the current `SSLProtocol` implementation
is prepared to support it.


I plan to use a switch between optimized socket access for
"normal" connections (it might be necessary to exclude `uvloop` as well)
and standard socket access for SSL connections (and maybe `uvloop`).
This should allow to run the complete test suite and analyse test
failures.

Once all tests pass, we can compare the speed of
"normal" and "optimized" soccket access and determine
whether the additional complexity (and risk) is worth the effort.

d-maurer:
Dieter Maurer wrote at 2023-1-26 23:19 +0100:
>Kirill Smelkov wrote at 2023-1-26 13:27 -0800:
>> ...
>>This way maybe if we disable re-negotiation (`OP_NO_RENEGOTIATION`) we can count on write never triggering a need for read. After all TLS Record Protocol is very simple and in the protocol itself there is no need to want to read if one side does write at application level.

I got an idea how to make the approach safe for SSL connections, too:

  * Replace the write lock by a loop lock.

  * The loop's internal method `_run_once` (it performs one loop run)
    is overridden to hold the loop lock during its execution

  * While an application thread executes code originally in the
    IO thread, it holds the loop lock

This guarantees that IO and application code is never executed
concurrently. Application code can only break in between loop runs --
at a moment when the internal IO state is safe. The application
code does only things usually done in loop callbacks; after it has
finished, the state is safe again.

As the application code is still run in a different thread, it still
cannot use standard `asyncio` features (e.g. `call_soon`, `call_later`, ...).
Thus, breakage is still possible (should the implementation
use such features. Because we know that the loop is passive
when the loop lock is held, we might be able to take over the loop
into the application thread.

The approach requires the override of an internal loop method.
The might be impossible for loops implemented in "C" (e.g. `uvloop`).

navytux:
Dieter, I agree - it is better to first implement simple non-SSL case to see what can be achieved with this approach. By the way the other thing I noticed about SSL is that in Python 3.11 the SSLProtocol no longer does `resume_read` on write codepath. Mentining this just for the reference as, once again, you are right that we should first see about what are the RTT latency improvements if we do it for regular non-SSL sockets.

navytux:
And the lock taken by `_run_once` - while correct might turn out to be too coarse killing any potential improvement because

1. `_run_once` is run constantly e.g. by `run_forever` in a `while 1` loop,
2. now client thread, trying to avoid switching to IO thread, would need for the same IO thread to release the lock.

But maybe I'm wrong on this.

navytux:
P.S. I would suggest to do a simple test first with simple asyncio client and echo server. With first client doing requests, similarly to ZEO, via `call_soon_threadsafe`, and in another version - by explicitly locking tx and transmitting directly without changing thread, then waiting for IO thread to receive the result.

Then, when we have the number for how RTT improves and via which way, we could start changing ZEO with more confidence.

d-maurer:
Kirill Smelkov wrote at 2023-1-27 12:59 -0800:
> ...
>1. `_run_once` is run constantly e.g. by `run_forever` in a `while 1` loop,

You are right: `_run_once` includes the waiting for relevent
events. What needs to be lock protected would be the part
of `_run_once` starting with `_process_events` until the end of
`_run_once`.

This leaves the `_scheduled` processing before `select` unprotected,
but this is used only by `call_later` which can only be used inside
the IO thread.

>2. now client thread, trying to avoid switching to IO thread, would need for the same IO thread to release the lock.

`_run_once` spents most of its time in waiting for relevant events.
Assuming that most event processing is short there is a good chance
for no thread switch during this processing. This implies:
when application code runs, it is likely that the IO thread is
waiting for events.

d-maurer:
I have run `zodbshootout` and found significant speedup:
```markdown
| Benchmark                                                    | master | with_write_lock      |
|--------------------------------------------------------------|:------:|:--------------------:|
| {c=1 threads, o=1000} cs: add 1000 objects                   | 450 ms | 225 ms: 2.00x faster |
| {c=1 threads, o=1000} cs: store 1000 raw pickles             | 431 ms | 194 ms: 2.22x faster |
| {c=1 threads, o=1000} cs: update 1000 objects                | 477 ms | 239 ms: 1.99x faster |
| {c=1 threads, o=1000} cs: read 1000 cold objects             | 454 ms | 348 ms: 1.30x faster |
| {c=1 threads, o=1000} cs: read 1000 cold prefeteched objects | 173 ms | 244 ms: 1.41x slower |
| {c=1 threads, o=1000} cs: write/read 1000 objects            | 894 ms | 573 ms: 1.56x faster |
| Geometric mean                                               | (ref)  | 1.53x faster         |
```
The optimization is not possible for `uvloop`: its transport does not allow to override `write`.
The optimization is not implemented for SSL access - its code looks too complicated to risk the monkey patch.

d-maurer:
This PR and #225 achieve the same degree of speed improvements but #225 is significantly simpler. Therefore, I give it up in favor of #225.

PR #225, opened 2023-02-02
Optimize via direct socket access with loop lock concurrency protection

d-maurer:
Alternative (wrt #224) optimization following @navytux idea to access the socket directly.
While #224 monkey patches the write operation to include a synchronizing lock, this PR monkey patches the loop and includes a loop lock which ensures that an application thread cannot execute a task step while the loop processes callbacks.

Initially, I had hoped that this approach would work for SSL connections, too. But the SSL writes include non thread safe loop operations. Thus, like for #224, SSL access cannot be optimized.

Nevertheless, this approach is simpler and therefore potentially more reliable.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,13 @@ Changelog
   ``password`` - support for such basic auth was dropped in 2016 before ZEO 5.0
   was released.
 
+- Optimize ZEO server interactions in typical cases
+  (e.g. no SSL, no ``uvloop``) by allowing client application
+  threads to access the server communication socket directly.
+  This can speed up ZEO operations considerable, especially reads.
+  For details see
+  `#225 <https://github.com/zopefoundation/ZEO/pull/225>`_.
+
 
 5.4.0 (2023-01-18)
 ------------------
--- a/src/ZEO/asyncio/_futures.pyx
+++ b/src/ZEO/asyncio/_futures.pyx
@@ -218,6 +218,7 @@ cdef class CoroutineExecutor:
     cdef object awaiting  # future we are waiting for
     cdef bint   cancel_requested  # request to be canceled from .cancel
     cdef object cancel_msg  # ^^^ cancellation message
+    cdef object lock  # ``None`` or the loop lock
 
     def __init__(self, task, coro):
         """execute *coro* on behalf of *task*."""
@@ -226,8 +227,19 @@ cdef class CoroutineExecutor:
         self.awaiting = None
         self.cancel_requested = False
         self.cancel_msg = None
+        self.lock = None
 
     cpdef step(self):
+        # If ``self.lock`` is not ``None``, we may run in an
+        # application thread; acquire the lock to prevent
+        # concurrency with the IO thread
+        if self.lock is None:
+            self._step()
+        else:
+            with self.lock:
+                self._step()
+            
+    cdef _step(self):
         await_result = None  # with what to wakeup suspended await
         await_resexc = False # is it exception?
         awaiting = self.awaiting
@@ -356,10 +368,10 @@ cdef class ConcurrentTask(ConcurrentFuture):
     def __init__(self, coro, loop):
         ConcurrentFuture.__init__(self, loop=loop)
         self.executor = CoroutineExecutor(self, coro)  # reference cycle
-        loop.call_soon_threadsafe(self._start)
+        self._start()
 
-    cpdef _start(self):
-        self.executor.step()
+    cdef _start(self):
+        self._loop.call_soon_threadsafe(self.executor.step)
 
     cpdef cancel(self, msg=None):
         """external cancel request.
@@ -374,7 +386,13 @@ cdef class ConcurrentTask(ConcurrentFuture):
         return ConcurrentFuture.cancel(self, msg)
 
 
-# @coroutine - only py implementtion
+cdef class FastConcurrentTask(ConcurrentTask):
+    """A oncurrent task with the first step run by the application thread.
 
+    Subsequent steps are tpyically run by the IO thread but can in
+    rare cases also be run by the application thread.
+    """
 
-run_coroutine_threadsafe = ConcurrentTask
+    cdef _start(self):
+        self.executor.lock = self._loop.lock
+        self.executor.step()
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -65,6 +65,7 @@ import sys
 import threading
 from asyncio import CancelledError
 from asyncio import TimeoutError
+from itertools import count
 
 import ZODB.event
 import ZODB.POSException
@@ -1069,6 +1070,7 @@ class ClientRunner:
         if self.__closed:
             return
         self.__closed = True
+        logger.debug("closing")
         call = self.io_call
         self.io_call = self._io_call_after_closure
         loop = self.loop
@@ -1124,10 +1126,11 @@ class ClientThread(ClientRunner):
                          )
         self.thread = threading.Thread(
             target=self.run_io_thread,
-            name="%s zeo client networking thread" % client.__name__,
-            )
+            name="%s zeo client networking thread %s"
+            % (client.__name__, make_thread_id()))
         self.thread.daemon = True
         self.started = threading.Event()
+        logger.debug("starting %s", self.thread.name)
         self.thread.start()
         self.started.wait()
         if self.exception:
@@ -1165,6 +1168,7 @@ class ClientThread(ClientRunner):
         """
         if not self.__closed:
             self.__closed = True
+            logger.debug("close requested")
             loop = self.loop
             if loop is None:  # pragma no cover
                 # we have never been connected
@@ -1192,3 +1196,12 @@ def cancel_task(task):
     if sys.version_info < (3,9):
         task.add_done_callback(
             lambda future: future.cancelled() or future.exception())
+
+
+# Debugging aid
+thread_counter = count()
+
+def make_thread_id():
+    """return new unique thread id -- for log analysis."""
+    return next(thread_counter)
+
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -65,6 +65,7 @@ import sys
 import threading
 from asyncio import CancelledError
 from asyncio import TimeoutError
+from asyncio import get_event_loop
 from itertools import count
 
 import ZODB.event
@@ -524,14 +525,16 @@ class ClientIO:
         # loop operation in ``_write_app_data``.
         self.direct_socket_access = \
               ssl is None and \
-              hasattr(loop, "_process_events") and \
-              hasattr(loop, "_run_once")
+              all(hasattr(loop, m) for m in
+                  ("_process_events", "_run_once",
+                   "_write_to_self", "_add_writer"))
         if self.direct_socket_access and not hasattr(loop, "lock"):
             # direct socket access possible
             # patch the loop
             loop.lock = threading.RLock()
             ori_process_events = loop._process_events
             ori_run_once = loop._run_once
+            ori_add_writer = loop._add_writer
 
             def _process_events(*args):
                 loop.lock.acquire()
@@ -541,8 +544,18 @@ class ClientIO:
                 ori_run_once()
                 loop.lock.release()
 
+            def _add_writer(*args):
+                ori_add_writer(*args)
+                try:
+                    c_loop = get_event_loop()
+                except Exception:
+                    c_loop = None
+                if c_loop is not loop:
+                    loop._write_to_self()
+
             loop._process_events = _process_events
             loop._run_once = _run_once
+            loop._add_writer = _add_writer
             logger.info("loop patched to allow direct socket access")
 
         self.disconnected(None)
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -5,10 +5,19 @@ The client interface implementation is split into two parts:
 ``ClientRunner`` methods are executed in the calling thread,
 ``ClientIO`` methods in an ``asyncio`` context.
 
-``ClientRunner`` calls ``ClientIO`` methods indirectly via
-either ``loop.call_soon_threadsafe`` (the async calls
-``call_async`` and ``call_async_iter``) or
-``run_coroutine_threadsafe`` (methods with suffix ``_co``).
+``ClientRunner`` calls ``ClientIO`` methods in three different ways:
+
+  directly
+     the async calls ``call_async`` and ``call_async_iter``
+
+  via ``run_coroutine_threadsafe``
+     the methods with ``_co`` suffix
+
+  via ``run_coroutine_fast``
+     the methods with ``_fco`` suffix.
+     Note that those methods are partially run in the application thread
+     and therefore cannot use standard ``asyncio`` functionality.
+
 ``ClientIO`` does not call ``ClientRunner`` methods; however, it
 can call ``ClientStorage`` and ``ClientCache`` methods.
 Those methods must be thread safe.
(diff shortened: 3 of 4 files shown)


d-maurer:
@navytux The strange test behavior is partly understood: your (relatively new) test `check_race_external_invalidate_vs_disconnect` occasionally leaves threads behind. Some of those threads are not "daemonic", as a consequence the interpreter shutdown is delayed until those threads stop -- what apparently never happens.

Not yet understood: why GHA shows the output only after the test run is cancelled.

@navytux Would you check your test? At least, all used threads should be "daemonic" -- in order not to interfere with the shutdown. In addition, it might be adequate to let the test fail when the threads are not shut down properly. The same problem (use of non daemonic threads) might affect other of your tests as well.

In my local setup, `zope-testrunner -f --test-path ../ZEO/src -s ZEO -a 1000 -vv -t "external_invalidate_vs_disconnect.*Mapping"` often leaves threads behind and then blocks in the shutdown.

I will try to find out why the threads do not reliably terminate in this branch.

navytux:
Thanks, Dieter. I will try to look into the test in some time. It is again, sadly, a difficult time here and I'm overloaded. Hopefully it might get better in some time. I appologize for the inconvenience.

P.S. One thing I thought out after seeing https://github.com/zopefoundation/ZEO/pull/224#issuecomment-1410125616 is this: given the speedup is significant I've thought about the whole thing as follows: on Linux threads are full processes. It is just that threads inside the same "process" are processes that share address space. This way before #224 what we have for typical `loadBefore` is 4 switches in between processes:

1. Client thread -> IO thread  via `call_soon_threadsafe` which internally sends wakeup via internal socket.
2. IO thread -> ZEO server process (request)
3. ZEO server -> IO thread (reply)
4. IO thread -> client thread (result)

#224 removes 1, but leaves 2,3 and 4. And even if we remove 1 process switch out of 4 the improvements are significant.

This way it makes sense to see if we can remove all but absolutely necessary processes switches. Looking at the problem from this point of view the other unnecessary switch is 3: we could send directly request to ZEO server from Client thread, and we could receive ZEO server reply also directly by that same client thread avoiding intermediate point on the IO thread.

That would work straightforwardly if we have only one client thread, but the reason IO thread exists at all is because we can have several client threads each issuing requests simultaneously.

So what I thought is this: we can organize execution of asyncio loop in our own LoopExecutor. Then when a client wants to issue server request, it creates corresponding coroutine, and then does the following: check if the loop is already running. If not - start running the loop directly by that client thread  until that loading coroutine completes (and gives the result). If loop is already running - schedule coroutine via call_soon_threadsafe and waits the result via ConcurrentFuture. This way at least one client thread would be able to issue server request directly and receive it also directly also serving other client threads along the way the same way as IO thread currently does. And also we will need to keep on running the IO thread, that would periodically run the same procedure to run the loop, so that from-server invalidations are still received even if there is no client activity. All the above logic could be put into LoopExecutor, and then every place that does the request does just create coro, then loop_exe.run_coro(coro).

The loop will be run at arbitrary thread, but only at one thread at a time. I've quickly checked what is the usage of threading.local in asyncio implementation and I did not see any place which might prevent this scheme to work.

I don't have numbers, but I believe reducing the number of process switches from 3 to 2 could bring even more significant latency reduction. That's why I would consider trying this scheme.

Kirill

d-maurer:
Kirill Smelkov wrote at 2023-2-2 14:10 -0800:
> ...
>P.S. One thing I thought out after seeing https://github.com/zopefoundation/ZEO/pull/224#issuecomment-1410125616 is this: given the speedup is significant I've thought about the whole thing as follows: on Linux threads are full processes. It is just that threads inside the same "process" are processes that share address space. This way before #224 what we have for typical `loadBefore` is 4 switches in between processes:
>
>1. Client thread -> IO thread  via `call_soon_threadsafe` which internally sends wakeup via internal socket.
>2. IO thread -> ZEO server process (request)
>3. ZEO server -> IO thread (reply)
>4. IO thread -> client thread (result)
>
>#224 removes 1, but leaves 2,3 and 4. And even if we remove 1 process switch out of 4 the improvements are significant.
>
>This way it makes sense to see if we can remove all but absolutely necessary processes switches. Looking at the problem from this point of view the other unnecessary switch is 3: we could send directly request to ZEO server from Client thread, and we could receive ZEO server reply also directly by that same client thread avoiding intermediate point on the IO thread.

Possible in principle but not easy to achieve.
You would make the ZEO communication synchronous: the application
thread sends the request and waits for the result.
This would work with RPC calls **BUT** the server can also send
asynchronous calls **AND** you must guarantee that messages
are processed in sent order.

A further difficulty: long running requests (such as pack).
You would not want that during such calls other requests are blocked.
This means that you cannot use simple synchronous communication;
instead the communication channel must be multiplexed between
different application threads and an additional actor processing
asnychonous messages.

> ...
>So what I thought is this: we can organize execution of asyncio loop in our own LoopExecutor. Then when a client wants to issue server request, it creates corresponding coroutine, and then does the following: check if the loop is already running.

The loop typically synchronously waits in a `select/poll`.
When it does that, it must have a thread which does the waiting.
As you have pointed out for my "lock `_run_once`" idea, the
loop is essentially lways running -- the run just consists in
waits to a large part.

Remember also that the loop must always be prepared to
process asynchronus messages -- even when no application
thread is interested in server interaction.

d-maurer:
`zodbshootout` (`zodbshootout --pack -c 1 --fast -l 5 --include-mapping no --threads shared shootout.conf add store update cold prefetch_cold warm`) comparison of `master`, #224 and #225:
```markdown
(py310) pyperf compare_to --table --table-format=md master with_write_lock with_loop_lock
| Benchmark                                                    | master | with_write_lock      | with_loop_lock       |
|--------------------------------------------------------------|:------:|:--------------------:|:--------------------:|
| {c=1 threads, o=1000} cs: add 1000 objects                   | 450 ms | 225 ms: 2.00x faster | 224 ms: 2.01x faster |
| {c=1 threads, o=1000} cs: store 1000 raw pickles             | 431 ms | 194 ms: 2.22x faster | 196 ms: 2.20x faster |
| {c=1 threads, o=1000} cs: update 1000 objects                | 477 ms | 239 ms: 1.99x faster | 239 ms: 2.00x faster |
| {c=1 threads, o=1000} cs: read 1000 cold objects             | 454 ms | 348 ms: 1.30x faster | 337 ms: 1.34x faster |
| {c=1 threads, o=1000} cs: read 1000 cold prefeteched objects | 173 ms | 244 ms: 1.41x slower | 176 ms: 1.01x slower |
| {c=1 threads, o=1000} cs: write/read 1000 objects            | 894 ms | 573 ms: 1.56x faster | 565 ms: 1.58x faster |
| Geometric mean                                               | (ref)  | 1.53x faster         | 1.63x faster         |
```
The time for #224 "cold prefetched" is too high in the table above: it likely is comparable to that for #225.

Overall, #225 and #224 achieve the same speed improvements. Because #225 is simpler, I discontinue the #224 experiment.

navytux:
> You would make the ZEO communication synchronous

Dieter, thanks for feedback. While I might be missing something for sure it looks like I could not describe what I meant clearly and you misunderstood it a bit: I did not meant for ZEO client to become synchronous. It is still the same asyncio and the same loop. It is just that we run that loop in piece by thread1, then in piece by thread2, ... The loop is still run and runs _all_ coroutines registered to it. But each portion of the run can be handled by different runner thread. For a loadBefore we know that client thread is waiting for the result - thus we can delve into running the loop until the result is ready. And while the loop is running - it runs all other coroutines as well.

navytux:
> Remember also that the loop must always be prepared to
> process asynchronus messages -- even when no application
> thread is interested in server interaction.

For this IO thread periodically tries to run the loop with low priority - so that asynchronous invalidation are processed irregardless of whether there is client activity or not.

navytux:
> @navytux Would you check your test?

@d-maurer, I've looked at the test quickly, and besides the improvement you did in https://github.com/zopefoundation/ZODB/pull/376 to explicitly fail if thread join does not succeeds, I do not see a dedlock potential in the test logic itself.

Yes, I looked only briefly, and might have missed something, but to me the deadlock caught here (yes, I also get it on your branch with `zope-testrunner -f --test-path ../ZEO/src -s ZEO -a 1000 -vv -t "external_invalidate_vs_disconnect.*Mapping"`  while on ZEO master there is no such thing) suggests to me that there is likely some concurrency bug added in hereby PR.

Would you please check it further what is going on? If you believe everything is right on your side, I could probably try to check it by myself but unfortunately only in some time from now.

d-maurer:
Kirill Smelkov wrote at 2023-2-3 11:51 -0800:
> ...
>For this IO thread periodically tries to run the loop with low priority

Python thread do not have a priority.

>- so that asynchronous invalidation are processed irregardless of whether there is client activity or not.

For the correctness, it is vital that messages are processed
in sent order. You cannot easily delegate asynchronous calls
to a low priority activity.

d-maurer:
Kirill Smelkov wrote at 2023-2-3 12:20 -0800:
> ...
>@d-maurer, I've looked at the test quickly, and besides the improvement you did in https://github.com/zopefoundation/ZODB/pull/376 to explicitly fail if thread join does not succeeds, I do not see a dedlock potential in the test logic itself.

I am investigating.

Unfortunately, the mass operations produced by this test
make the analysis difficult.
I have the following plan:

 1. Try to make the test more deterministic by setting the `random`
    seed

    While now different threads starve, I hope that
    after this the starving becomes more predictable.

 2. Introduce log messages in the test to be able to
    determine which operations do not finish from
    a log file analysis

 3. Once the operation is identified, add log messages to `ClientIO`
    to trace the execution of this operation.

>Yes, I looked only briefly, and might have missed something, but to me the deadlock caught here (yes, I also get it on your branch with `zope-testrunner -f --test-path ../ZEO/src -s ZEO -a 1000 -vv -t "external_invalidate_vs_disconnect.*Mapping"`  while on ZEO master there is no such thing) suggests to me that there is likely some concurrency bug added in hereby PR.

This is well possible.

As stressed several times before, we are doing potentialy
dangerous things: the `asyncio` code (loop, transports, protocols, ...)
has been designed under the assumption that it is run in a single thread.
We are breaking this assumption running part of the code in
applications threads.

I think that the affected code is never run concurrently
but this may not be enough to guarantee correctness.
One example to the contrary has been the SSL transport:
it contains code which must be run by the IO thread
and checks this; other parts, too, might require this but
make no check for it. Then bad things could happen - e.g. the loss
of events and a potential deadlock.


>Would you please check it further what is going on?

Sure.

>If you believe everything is right on your side, I could probably try to check it by myself but unfortunately only in some time from now.

You are an expert for the analysis of race conditions.
Working together might give earlier results.

d-maurer:
Dieter Maurer wrote at 2023-2-3 22:29 +0100:
>Kirill Smelkov wrote at 2023-2-3 12:20 -0800:
> ...
>I think that the affected code is never run concurrently
>but this may not be enough to guarantee correctness.
>One example to the contrary has been the SSL transport:
>it contains code which must be run by the IO thread
>and checks this; other parts, too, might require this but
>make no check for it. Then bad things could happen - e.g. the loss
>of events and a potential deadlock.

@navytux
I have identified one of those locations: `loop._add_writer`.

`_add_writer` is called when an `asyncio` transport could not send
all data in its `write` call (and the output buffer was initially empty).
In this case, the remaining data is buffered and `_add_writer` called.
`add_writer` registers a `write_ready` event handler for the socket with
the event loop. This event handler should drain the output buffer
when the socket becomes write ready.

This works in the IO thread because `Transport.write` is called only
in callbacks. Therefore, the IO thread is never waiting for events
(in `select/poll`) and the following `select/poll` call will honor
the new event handler registration.

In our situation, however, `Transport.write` can be called by an
application thread. Then. there is a high change that the IO thread
is waiting in `select/poll`. This call will not see the new
event handler, potentially registered via `_add_writer` -- only
the following `select/poll` call.
As a consequence, the socket's `write_ready` event may remain undetected
leading to a deadlock.


With ZEO, the event loop should always have the "heartbeat" timeout
handler. This means that the `select/poll` should be interrupted
at least once in a minute and we should not see a real deadlock
(only a delay for up to 1 minute)
The observed test failure, however, indicates a real deadlock
(broken only when the server shuts down -- the corresponding EOF
is still seen).

This means that there are potentially further locations
which rely on being run by the IO thread -- and not an application
thread.


I will append a `loop._write_to_self()` to all situations in which
an application thread may run IO code.
This fixes the deadlock szenario described above --
but might not be enough for further similar cases
and potentially the case encountered by the failing test.

d-maurer:
Dieter Maurer wrote at 2023-2-4 09:26 +0100:
> ...
>@navytux
>I have identified one of those locations: `loop._add_writer`.
> ...
>With ZEO, the event loop should always have the "heartbeat" timeout
>handler. This means that the `select/poll` should be interrupted
>at least once in a minute and we should not see a real deadlock
>(only a delay for up to 1 minute)
>The observed test failure, however, indicates a real deadlock
>(broken only when the server shuts down -- the corresponding EOF
>is still seen).

As expected, the hole in `loop._add_writer` (while
indoubtedly problematic) has not been the cause of the
`check_race_external_invalidate_vs_disconnect` failure:
this was caused by a concurrency between the application
thread and the IO thread leaving a future in an inconsistent state:
I had protected only the coroutine interaction instead of the
whole coroutine executor step.


All tests are green now.

However, the `py37-zodbmaster` tests have initially failed.
The error pattern points towards a timing problem for the
`check_race_...` tests:
for tests, the ZEO server has a suicide timeout;
apparently, it killed itself before the test completed
(likely due to a high load on the testing infrastructure);
consequence: hundreads of `Connection lost` errors.
Maybe, we want to do something about this problem -- or
expect occasional failures.

d-maurer:
Final `zodbshootout` timings (`zodbshootout --pack --fast -l 5 --include-mapping no --threads shared shootout.conf add store update cold prefetch_cold warm`:
```markdown
| Benchmark                                                    | master      | PR                     | cython PR                 |
|--------------------------------------------------------------|:-----------:|:----------------------:|:--------------------------:|
| {c=2 threads, o=1000} cs: add 1000 objects                   | 894 ms      | 700 ms: 1.28x faster   | 435 ms: 2.05x faster       |
| {c=2 threads, o=1000} cs: store 1000 raw pickles             | 439 ms      | 352 ms: 1.25x faster   | 199 ms: 2.21x faster       |
| {c=2 threads, o=1000} cs: update 1000 objects                | 953 ms      | 787 ms: 1.21x faster   | 597 ms: 1.59x faster       |
| {c=2 threads, o=1000} cs: read 1000 cold objects             | 744 ms      | 524 ms: 1.42x faster   | 493 ms: 1.51x faster       |
| {c=2 threads, o=1000} cs: read 1000 cold prefeteched objects | 335 ms      | 344 ms: 1.03x slower   | 320 ms: 1.04x faster       |
| {c=2 threads, o=1000} cs: write/read 1000 objects            | 1.36 sec    | 1.23 sec: 1.10x faster | 1.19 sec: 1.14x faster     |
| Geometric mean                                               | (ref)       | 1.20x faster           | 1.53x faster               |
```
Unlike previous timings, those above use concurrency 2 (--> `c=2`) which is more realistic for high load than the previous concurrency 1. As a consequence, the timings cannot be compared.

Unfortunately, the speedup dropped considerably. Not sure, whether this is due to the increased concurrency or the stricter protection.

d-maurer:
Corresponding timings with concurrency 1 (`zodbshootout --pack -c 1 --fast -l 5 --include-mapping no --threads shared shootout.conf add store update cold prefetch_cold warm`):
```markdown
| Benchmark                                                    | master  | PR                   | cython PR              |
|--------------------------------------------------------------|:-------:|:--------------------:|:----------------------:|
| {c=1 threads, o=1000} cs: add 1000 objects                   | 461 ms  | 225 ms: 2.05x faster | 227 ms: 2.03x faster   |
| {c=1 threads, o=1000} cs: store 1000 raw pickles             | 431 ms  | 196 ms: 2.19x faster | 195 ms: 2.21x faster   |
| {c=1 threads, o=1000} cs: update 1000 objects                | 480 ms  | 362 ms: 1.33x faster | 240 ms: 2.00x faster   |
| {c=1 threads, o=1000} cs: read 1000 cold objects             | 455 ms  | 339 ms: 1.34x faster | 326 ms: 1.39x faster   |
| {c=1 threads, o=1000} cs: read 1000 cold prefeteched objects | 174 ms  | 175 ms: 1.01x slower | 165 ms: 1.05x faster   |
| {c=1 threads, o=1000} cs: write/read 1000 objects            | 891 ms  | 708 ms: 1.26x faster | 550 ms: 1.62x faster   |
| Geometric mean                                               | (ref)   | 1.47x faster         | 1.66x faster           |
```
Apparently, the efficiency of the optimization drops significantly with an increasing concurrency. The drop is more pronounced without the `cython` optimization.

navytux:
Dieter, is it ok if I take a bit of time break before reviewing this?

P.S. it would be also good to know for completeness whether https://github.com/zopefoundation/ZEO/pull/224 drops performance in similar way for `c=2`.

d-maurer:
Kirill Smelkov wrote at 2023-2-8 23:31 -0800:
>Dieter, is it ok if I take a bit of time break before reviewing this?

"A bit" is never a problem.

This PR is small; thus, hopefully, the break will not be in the order
of many months.

>P.S. it would be also good to know for completeness whether https://github.com/zopefoundation/ZEO/pull/224 drops performance in similar way for `c=2`.

I think this is very likely.

You can find out by running the `zodbshootout` performance measurement.

Note that #224 does not yet contain the `loop._add_writer` enhancement;
this may cause (apparently non deterministic) delays of up to 1 minute.


Taking `zodbshootout` measurements in a different environment may be good
in its own -- to get hints towards the influence of
my particular architecture.

navytux:
Thanks, Dieter. Yes, hopefully, the delay should not be several months, like it was with huge #195. And ok - I will run the benchmarks myself while reviewing this.

d-maurer:
Dieter Maurer wrote at 2023-2-9 08:49 +0100:
>Kirill Smelkov wrote at 2023-2-8 23:31 -0800:
> ...
>>P.S. it would be also good to know for completeness whether https://github.com/zopefoundation/ZEO/pull/224 drops performance in similar way for `c=2`.
>
>I think this is very likely.
>
>You can find out by running the `zodbshootout` performance measurement.
>
>Note that #224 does not yet contain the `loop._add_writer` enhancement;
>this may cause (apparently non deterministic) delays of up to 1 minute.

Note as well that I have not worked on `_futures.pyx`
for #224 -- it is still for `master` (and therefore lacks the required
locking).

d-maurer:
I have waited very long for positive review. In addition, @navytux confirmed in "https://github.com/zopefoundation/ZEO/pull/232#issuecomment-1659021222" another idea which -- under optimal condition -- could save not only one but 2 (out of 4) context switches. Therefore, I feel it likely that a positive review will never come and I close this PR.

Issue #226, opened 2023-03-10
Client does not retry failed connections


azmeuk:
In my development environment, I use a script to run a zeo server and a flask application configured as a zeo client. The connection between the client and the server is done with a unix socket.
Depending on the runs, the server may or may not be ready before the client is. With ZEO 5.4.0, when the client is ready first, it prints the following error message, and does not retry to connect:

```
2023-03-10 09:34:34,836:INFO:ZEO.asyncio.client:Connection to '/path/to/zeo.sock' failed, [Errno 2] No such file or directory
```

With ZEO 5.3.0 the client would have tried several times until the connection succeeds, which was very convenient in such situations.

I suggest implementing a timeout or a try again mechanism for client connection. What do you think?

azmeuk:
Installing the `uvloop` package solves the situation.

d-maurer:
Éloi Rivard wrote at 2023-3-10 07:30 -0800:
> ...
>Thank you for your answer. The error seems to be around `yield asyncio.sleep`, as the *retry connecting* message is never shown.

You might want to learn how long the sleep should be.
Usually, it would be between 1 and 2 seconds.
>
>Actually, if I wait long enough, I get this error message:
>
>```python
> ...
>  File "/path/to/ZEO/asyncio/client.py", line 706, in await_operational_co
>    raise ClientDisconnected("timed out waiting for connection")
>ZEO.Exceptions.ClientDisconnected: timed out waiting for connection
>2023-03-10 15:34:55,902:ERROR:asyncio:AsyncTask exception was never retrieved

This is the normal 30 s connection timeout.
The `sleep` time should be significantly shorter.

>future: <ZEO.asyncio.futures.AsyncTask object at 0x7f777695ea40>
>Traceback (most recent call last):
> ...
>    loop = events.get_running_loop()
>RuntimeError: no running event loop

Once the connection timeout occurred, the client is closed.
Part of this process is cancelling the connection process
and shutting down the loop.
Apparently, the loop is shut down before the cancelation is finished.
This causes the observed exception.

I agree with you that something does not work as expected with
the `sleep`.
At your place, I would try to find out how long the sleep should be
(e.g. by modifying the code and logging the sleep time).
If the sleep takes significantly longer that this time,
I would contact @navytux.
If the sleep time is much higher than the expected 1 to 2 seconds,
something went wrong with the parameter passing or the
`local_random` behaves strangely.

d-maurer:
[...]

PR #227, opened 2023-03-15
[ZEO5] asyncio.client: Fix connect logic wrt IO loop startup

navytux:
Éloi Rivard reports that starting from ZEO 5.4.0 client no longer retries failed connection to server via unix sockets.

That happens to be due to that `connect` coroutine starts to be executed before `loop.run_forever` is called and so `asyncio.sleep()` raises "no running event loop". Here are the details:

with the following debug patch:

```diff
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -145,6 +145,9 @@ def pop_futures(self):
         return futures

     def connect(self):
+        trace('connect')
+        import traceback as tb; tb.print_stack()
+        trace()
         if isinstance(self.addr, tuple):
             host, port = self.addr
             cr = lambda: self.loop.create_connection(  # noqa: E731
@@ -156,6 +159,9 @@ def connect(self):

         @coroutine
         def connect():
+            trace('coro connect')
+            import traceback as tb; tb.print_stack()
+            trace('running loop:', asyncio.events.get_running_loop())
             while not self.closed:
                 try:
                     yield cr()
@@ -1035,9 +1041,14 @@ def run_io_thread(self):
         try:
             loop = self.loop if self.loop is not None else new_event_loop()
             asyncio.set_event_loop(loop)
+            trace('>>> setup_delegation')
             self.setup_delegation(loop)
+            trace('<<< setup_delegation')
             loop.call_soon(self.started.set)
+            import time; time.sleep(3)
+            trace('>>> loop.run_forever')
             loop_run_forever(loop)
+            trace('<<< loop.run_forever')
         except Exception as exc:
             logger.exception("Client thread")
             self.exception = exc
@@ -1088,3 +1099,8 @@ def cancel_task(task):
     if sys.version_info < (3,9):
         task.add_done_callback(
             lambda future: future.cancelled() or future.exception())
+
+
+import sys
+def trace(*argv):
+    print(*argv, file=sys.stderr, flush=True)
```

and the following test program

```py

import logging
logging.basicConfig(level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')

import ZEO
print('AAA')
zcli = ZEO.client('/path/to/zeo.sock')
print('BBB')
```

the output is like this:

```
(py3.venv) kirr@deca:~/src/wendelin/z/ZEO$ ../ZEO5/x.py 2>&1
AAA
2023-03-14 07:21:01,569 - INFO - /path/to/zeo.sock ClientStorage (pid=49949) created RW/normal for storage: '1'
2023-03-14 07:21:01,569 - INFO - created temporary cache file 3
2023-03-14 07:21:01,570 - DEBUG - Using selector: EpollSelector
>>> setup_delegation
2023-03-14 07:21:01,570 - DEBUG - disconnected <ZEO.asyncio.client.ClientIO object at 0x7f03c50077c0> None
2023-03-14 07:21:01,570 - DEBUG - try_connecting <ZEO.asyncio.client.ClientIO object at 0x7f03c50077c0>
connect
  File "/usr/lib/python3.9/threading.py", line 912, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 1045, in run_io_thread
    self.setup_delegation(loop)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 861, in setup_delegation
    self.client = client = ClientIO(loop, *self.__args, **self.__kwargs)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 477, in __init__
    self.disconnected(None)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 537, in disconnected
    self.try_connecting()
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 549, in try_connecting
    self.protocols = [
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 550, in <listcomp>
    Protocol(self.loop, addr, self,
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 93, in __init__
    self.connect()
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 149, in connect
    import traceback as tb; tb.print_stack()

coro connect
  File "/usr/lib/python3.9/threading.py", line 912, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 1045, in run_io_thread
    self.setup_delegation(loop)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 861, in setup_delegation
    self.client = client = ClientIO(loop, *self.__args, **self.__kwargs)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 477, in __init__
    self.disconnected(None)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 537, in disconnected
    self.try_connecting()
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 549, in try_connecting
    self.protocols = [
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 550, in <listcomp>
    Protocol(self.loop, addr, self,
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 93, in __init__
    self.connect()
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 178, in connect
    self._connecting = Task(connect(), self.loop)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/futures.py", line 390, in __init__
    self.executor.step()
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/futures.py", line 271, in step
    result = self.coro.send(await_result)
  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/client.py", line 163, in connect
    import traceback as tb; tb.print_stack()
<<< setup_delegation

>>> loop.run_forever
...
```

which shows that `connect` coroutine indeed starts to be executing immediately by `self._connecting = Task(connect(), self.loop)` because that Task is our task with immediate callback invocations. In other words here we try to run asyncio code outside of a running loop and that's why it breaks.

The fix is similar in style to d27a2cd1 and works by moving initial "connect" code to be executed from under running event loop.

For the reference: we do not have this problem on master, because there we changed `._connecting` to be native asyncio task instead of our own in 63429391
+ e3b8658d, and that has the side effect of not running connect coroutine immediately and instead scheduling it to be run when the loop is run. Still I would consider to apply my fix to master as well because it is potentially safer to do async-related activity from under already setup running loop.

/reported-by @azmeuk
/helped-by   @d-maurer
/fixes       https://github.com/zopefoundation/ZEO/issues/226

### DIFF
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -1035,8 +1035,12 @@ class ClientThread(ClientRunner):
         try:
             loop = self.loop if self.loop is not None else new_event_loop()
             asyncio.set_event_loop(loop)
-            self.setup_delegation(loop)
-            loop.call_soon(self.started.set)
+            # setup_delegation triggers to execute coroutines - we need to call
+            # it from under running loop
+            def _():
+                self.setup_delegation(loop)
+                self.started.set()
+            loop.call_soon(_)
             loop_run_forever(loop)
         except Exception as exc:
             logger.exception("Client thread")
--- a/src/ZEO/tests/ConnectionTests.py
+++ b/src/ZEO/tests/ConnectionTests.py
@@ -943,6 +943,30 @@ class ReconnectionTests(CommonSetupTearDown):
         logger.info("checkReconnection(): finished")
         self._storage.close()
 
+    def checkRetryConnectingUNIX(self):
+        # Check that the client retries connecting after failing initial
+        # attempt even on UNIX sockets: if zeo.sock does not exist at all,
+        # asyncio will fail "connecting" future without waiting for asyncio
+        # loop to start, and that was leading to crashes in connecting code
+        # because it was assuming it was running under asyncio loop.
+        # https://github.com/zopefoundation/ZEO/issues/226
+        self.shutdownServer()
+        self._servers = []
+        self.addr = ['/path/to/zeo.sock']
+
+        import zope.testing.loggingsupport
+        handler = zope.testing.loggingsupport.InstalledHandler(
+            'ZEO.asyncio.client')
+
+        self._storage = self.openClientStorage('test', 1000, wait=False)
+
+        time.sleep(2)
+        log = str(handler)
+        handler.uninstall()
+        self.assertIn("retry connecting", log)
+
+        self._storage.close()
+
     def checkMultipleServers(self):
         # Crude test-- just start two servers and do a commit at each one.
 


d-maurer (changes requested):
Earlier, I argued that the connection coroutine should not use our optimized `Task` but the usual `asyncio.Task` (i.e. it should get started via `asyncio.ensure_future`). My argument then: we do not know the implementation details of the connection process (a loop responsibility); it is well possible that this requires a full blown `Task` implementation - and not our optimized but limited implementation. I had thought that I had convinced you then.

If you implement the solution above now, it will solve the current problem (because the first step of a standard task is "called soon" rather than immediately -- i.e. delayed until the loop is running) and you are safer regarding potential task related requirements by the connection process.

d-maurer:
Kirill Smelkov wrote at 2023-3-30 11:59 -0700:
> ...
>I do not recall all the details now, but the reason we cannot do it for 5.x is that mixing our own AsyncTask and native asyncio/trollius Task does not play well with our CoroutineExecutor in 5.x where it manually implements `yield from` mechanics.

In fact, a `Task` is (more precisely: encapsulates) a
coroutine executor. Therefore, a native `Task` has nothing to do
with our coroutine executor -- it always uses its own one.

It is possible that the `connect` coroutine requires
a standard ***@***.***` decorator (instead of yours which
essentially only checks the type)
and that it must replace the `yield` in a way to be compatible
with a standard `Task` (when I remember right, this is `yield from`
in Python 3; no idea what it is for `trollius`).

> ...
> testClientBasics (ZEO.asyncio.tests.ClientTests)/usr/lib/python3.9/asyncio/events.py:80: RuntimeWarning: coroutine 'sleep' was never awaited

The reason might be the use of `yield` rather than `yield from`.

> ...
> test_multiple_addresses (ZEO.asyncio.tests.ClientTests)
>
>Error in test test_multiple_addresses (ZEO.asyncio.tests.ClientTests)
>Traceback (most recent call last):
>  File "/usr/lib/python3.9/unittest/case.py", line 59, in testPartExecutor
>    yield
>  File "/usr/lib/python3.9/unittest/case.py", line 593, in run
>    self._callTestMethod(testMethod)
>  File "/usr/lib/python3.9/unittest/case.py", line 550, in _callTestMethod
>    method()
>  File "/home/kirr/src/wendelin/z/ZEO/src/ZEO/asyncio/tests.py", line 528, in test_multiple_addresses
>    delay, func, args, handle = loop.later.pop(0)
>IndexError: pop from empty list

This error (and the following similar ones) may have the following
cause:
Our test loop uses `loop.later` to remember the `call_later` calls.
The `pop from empty list` indicates that at that point there was not yet such
a call. Because a standard `Task` delays each step executions
by one loop turn, it is well possible that the test requires adaptation
to give the task sufficient time for the execution.

To check whether this cause is really responsible for the error,
I suggest to introduce a small `sleep` into the test.
If this is the case, we could try later to replace the `sleep` again
with something not dependent on the timing of the test infrastructure.

navytux:
Let's go the way Dieter proposes in https://github.com/zopefoundation/ZEO/pull/228.

PR #228, opened 2023-04-09
Fix #226 using a standard `asyncio.Task` 

### LINKED ISSUE #226: Client does not retry failed connections

d-maurer:
Fixes #226.

This PR is an alternative to #227. Both fix #226.
While #227 fixes the immediate cause for the #226 failure, i.e. the connection process had made the implicit assumption that the loop was already running not fulfilled with ZEO, this PR approaches the problem in a more general way: the connection process is not implemented by ZEO but by foreign code; this code may expect a standard `Task`/coroutine executor, not the optimized but limited one of the ZEO implementation. For this reason, this PR uses a standard `asyncio.Task` to control the connection process.
A standard `Task` automatically ensures that its first step is only run when the loop is running.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@ Changelog
 5.4.1 (unreleased)
 ------------------
 
+- Fix `#226 <https://github.com/zopefoundation/ZEO/issues/226>`_
+  using an ``asyncio.Task`` (rather than a proprietary,
+  specially ZEO optimized task) for the connection process.
+  This is advisable because the connection process is not
+  implemented by ZEO but by foreign code which may
+  have problems with the limitations of ZEO tasks.
+
 
 5.4.0 (2023-01-18)
 ------------------
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -159,7 +159,11 @@ class Protocol(base.ZEOBaseProtocol):
             cr = lambda: self.loop.create_unix_connection(  # noqa: E731
                 self.protocol_factory, self.addr, ssl=self.ssl)
 
-        self._connecting = asyncio.Task(
+        # Usually, we use our optimized but feature limited tasks
+        # to run coroutines. Here we use a standard task
+        # because we do not know which task features the connect
+        # coroutine needs.
+        self._connecting = self.loop.create_task(
             connect_coroutine(cr, self, logger, local_random.random),
             loop=self.loop)
 
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -42,6 +42,11 @@ from .marshal import encoder, decoder
 from .futures import Future, AsyncTask as Task, \
      coroutine, return_, run_coroutine_threadsafe, switch_thread
 
+if sys.version_info[0] == 2:
+    from .connectco_2 import connect_coroutine
+else:
+    from .connectco_3 import connect_coroutine
+
 logger = logging.getLogger(__name__)
 
 Fallback = object()
@@ -154,22 +159,9 @@ class Protocol(base.ZEOBaseProtocol):
             cr = lambda: self.loop.create_unix_connection(  # noqa: E731
                 self.protocol_factory, self.addr, ssl=self.ssl)
 
-        @coroutine
-        def connect():
-            while not self.closed:
-                try:
-                    yield cr()
-                    return
-                except asyncio.CancelledError:
-                    logger.info("Connection to %r cancelled", self.addr)
-                    raise
-                except Exception as exc:
-                    logger.info("Connection to %r failed, %s",
-                                self.addr, exc)
-                yield asyncio.sleep(self.connect_poll + local_random.random())
-                logger.info("retry connecting %r", self.addr)
-
-        self._connecting = Task(connect(), self.loop)
+        self._connecting = asyncio.Task(
+            connect_coroutine(cr, self, logger, local_random.random),
+            loop=self.loop)
 
     def connection_made(self, transport):
         logger.debug('connection_made %s', self)
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -164,8 +164,7 @@ class Protocol(base.ZEOBaseProtocol):
         # because we do not know which task features the connect
         # coroutine needs.
         self._connecting = self.loop.create_task(
-            connect_coroutine(cr, self, logger, local_random.random),
-            loop=self.loop)
+            connect_coroutine(cr, self, logger, local_random.random))
 
     def connection_made(self, transport):
         logger.debug('connection_made %s', self)
--- a/src/ZEO/asyncio/connectco_2.py
+++ b/src/ZEO/asyncio/connectco_2.py
@@ -1,0 +1,19 @@
+"""Python 2 implementation for ``connect_coroutine``."""
+
+from .compat import asyncio
+
+@asyncio.coroutine
+def connect_coroutine(cr, self, logger, random):
+    while not self.closed:
+        try:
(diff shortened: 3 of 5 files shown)


navytux:
Dieter, thanks.

First of all I appologize for the delay with replying on my side wrt https://github.com/zopefoundation/ZEO/pull/227. And regarding hereby pull request I dislike a bit that we diverge the code for connect coroutine in between 5.x and master, but in the end it is one small change and does not look to be very important. I'm thus ok with your patch. The only amendment I would advise to make is to put a comment about _why_ `._connecting` is setup with native asyncio task instead of ours. I.e. just copy the following comment from master, where you already explained it well:

```diff
--- a/src/ZEO/asyncio/client.py
+++ b/src/ZEO/asyncio/client.py
@@ -159,6 +159,10 @@ def connect(self):
             cr = lambda: self.loop.create_unix_connection(  # noqa: E731
                 self.protocol_factory, self.addr, ssl=self.ssl)
 
+        # Usually, we use our optimized but feature limited tasks
+        # to run coroutines. Here we use a standard task
+        # because we do not know which task features the connect
+        # coroutine needs.
         self._connecting = asyncio.Task(
             connect_coroutine(cr, self, logger, local_random.random),
             loop=self.loop)
```

Thanks once again for making your take at this, and I'm sorry for the delays on my side.

Kirill

navytux (approved):

d-maurer:
Thank you for your feedback!

Kirill Smelkov wrote at 2023-4-10 08:46 -0700:
>First of all I appologize for the delay with replying on my side wrt https://github.com/zopefoundation/ZEO/pull/227.

Do not worry: I meanwhile recognized that my hypothesis there have been wrong.

>And regarding hereby pull request I dislike a bit that we diverge the code for connect coroutine in between 5.x and master, but in the end it is one small change and does not look to be very important.

The primary problem has been:
the modern `asyncio.Task` no longer supports `yield` based
generators but insists on `yield from` style.

My first attempt has been to define a `yield2yieldfrom` decorator
which emulates `yield from` style generators by `yield` generators
based on PEP 380.
Unfortunately, it hit the `return/StopIteration` incompatibility
between Python 2 and Python 3 (Python 2 does not allow `return XXX`
while Python 3 does not allow `raise StopIteration(...)` in a
generator).

In your 5.x work, you have used your own `_GenReturn` and `return_`
to work around this incompatibility.
Of course, `asyncio.Task` does not know about your `_GenReturn`;
but the `yield -> yield from` emulation could in principle
handle it.

I decided againt this approach because it is quite complex
and required only for this single place involving only
relatively simple logic.

>I'm thus ok with your patch.

If you like, we could look again at the alternative skechted above.

>The only amendment I would advise to make is to put a comment about _why_ `._connecting` is setup with native asyncio task instead of ours. I.e. just copy the following comment from master, where you already explained it well:

I have put the explanation into `CHANGES.rst`.
But I will follow your advice and put it also into the code.

navytux:
Dieter, thanks for feedback and you are welcome.

I agree the complexity is not worth unrolling it for this sole place. And we
already kind of already have `yield from` emulation in our CoroutineExecutor on
5.x .  It seems undesirable to duplicate that logic for only one particular
case. I would say either we use our own CoroutineExecutor and Task (and then
fix unix-socket issue via https://github.com/zopefoundation/ZEO/pull/227), or
we use native asyncio Task as you do here without unrolling additional
complexity.

I suggest we settle on the approach you presented here and move forward.

> I have put the explanation into `CHANGES.rst`.
> But I will follow your advice and put it also into the code.

Thanks.

Please feel free to merge your patch after this amendment.

Kirill

d-maurer:
@navytux Thank you for your review and feedback.

d-maurer:
Kirill Smelkov wrote at 2023-4-11 02:16 -0700:
> ...
>Please feel free to merge your patch after this amendment.

Apparently, we have a race condition in the check against
`zodbmaster`.

This time
("https://github.com/zopefoundation/ZEO/actions/runs/4666250704/jobs/8260629512?pr=228")
the `check_race_external_invalidate_vs_disconnect` fails;
initially it has been a different `check_race` test.

I do not think it has to do with my change.

@navytux Would you like to take a look?

d-maurer:
> @navytux Would you like to take a look?

I tried to analyze the `py310-zodbmaster` test failures. Unfortunately:
* no failure in my local environment
* the GH output is extremely hard to analyze. One important reason: the output from the hundreds of threads intermixes and it is almost impossible to associate the various output lines with the threads that produced them.

The following attachment contains the output of the first test failure (`BlobAdaptedFileStorageTests` `check_race_external_invalidate_vs_disconnect`).
[blob-adapted.txt](https://github.com/zopefoundation/ZEO/files/11208353/blob-adapted.txt)
@navytux Maybe you can find out what went wrong?

From my point of view, the tests must be significantly improved to get a chance to understand what happens.
At least, they must ensure that the exception output from the various threads does not get intermixed. Currently, this output is produced by `threading` which writes to the unbuffered `stderr`. To avoid mixing, the output could be created in the `Daemon` class, first formatted and then written via `os.write` directly to the file. Should this not yet prevent intermixing, a shared lock could be used.

Another improvement: when a test failure is detected, the ZEO server is shut down. This causes `ClientDisconnected` exceptions in all still active threads. The corresponding exceptions are not relevant; associated output should likely be avoided to facilitate the analysis of the initial problem.

@navytux Those are your tests. What do you think about the proposed improvements?

d-maurer:
> Another improvement:

And another improvement:
currently, we have
```
        for t in tg:
            t.join(60)

        if failed.is_set():
            self.fail('\n\n'.join([_ for _ in failure if _]))
```
This means: if we have both a verification failure (i.e. `failed.is_set`) and some thread does not stop, we lose the information about the verification failure. I suggest to catch a `t.join` exception, set `failed`, add a failure information to `failure` and continue the joins for remaining tasks with a significantly reduced timeout (e.g. 1 instead of 60).

d-maurer:
Kirill Smelkov wrote at 2023-4-14 05:14 -0700:
> ...
>My conclusion: this failure might have been due to server overload and test timing out.

Thank you for your feedback!

Indeed, I came to the same conclusion -- though by different means.
I noticed that the exception avalanche started about 1 min 28 s
after the layer start. I ran the layer tests locally with `-vvv`
and learned that all tests before `check_race...disconnect` where
fast and could explain about 28 s of the time above.
I adopted the hypothesis that the `check...disconnect` failure
could have been caused by `t.join(60)` requiring more than 1 minute.
This lets the test fail and leads to a server shutdown in the
`tearDown` -- causing huge numbers of secondary exceptions in the threads.
Because the failure message is not output immediately (but only after
`tearDown` and other runner activity), we can see some exception
messages from the threads before we see the failure message.

My work on "https://github.com/zopefoundation/ZODB/pull/380"
was partially motivated to clearly see when the test did not
finish within the expected time: one clear error message in this
case without the secondary messages.

My current hypothesis: the occasional test failures
come from special load situations. The motifications in
the ZODB PR above should allow us to check this hypothesis
(and give the test a bit more time, in this case).

navytux:
Dieter, thanks for feedback and sharing your thoughts.

PR #235, opened 2023-11-14
Add support for Python 3.12.

icemac:
Isort imports.


d-maurer (approved):

dataflake (approved):
not sure why the Pypy3 tests fail

d-maurer:
Jens Vagelpohl wrote at 2023-11-14 02:42 -0800:
>@dataflake approved this pull request.
>
>not sure why the Pypy3 tests fail

This looks like the effect of a `print` statement
(likely in server code).
Usually, the (doc)test is likely already finished before the print
is executed; here, a special load condition likely has let it
happen during the `ZEO.DB` call.

navytux:
@icemac, thanks for asking for review.

The first patch (`Add support for Python 3.12`) looks good to me.

The second patch (`Isort imports`): while I understand the rationale, when I look at the diff, for example at

```diff
-from ZODB.utils import p64, u64, z64, Lock, RLock
+from ZODB.utils import Lock
+from ZODB.utils import RLock
+from ZODB.utils import p64
+from ZODB.utils import u64
+from ZODB.utils import z64
```

to me it feels like the old version was better with higher signal to noise ratio.

But this is not a strong feeling and I won't object to this change if others find it convenient.

Kirill

icemac:
> not sure why the Pypy3 tests fail

PyPy3 prints some texts to stdout on GHA which break the doctests. I am going to ignore these texts.

icemac:
> But this is not a strong feeling and I won't object to this change if others find it convenient.

This is the default config for imports used in all other zopefoundation projects. We could change it here but I'd prefer not to do it.

icemac:

PR #237, opened 2023-11-21
Remove Python 2 leftovers.

### LINKED ISSUE #236: Look for more Python 2 leftovers
Additionally we get `DeprecationWarning: unittest.makeSuite() is deprecated and will be removed in Python 3.13. Please use unittest.TestLoader.loadTestsFromTestCase() instead.`

icemac:
Fixes #236.

### DIFF
--- a/src/ZEO/TransactionBuffer.py
+++ b/src/ZEO/TransactionBuffer.py
@@ -24,8 +24,7 @@ is used to store the data until a commit or abort.
 import tempfile
 
 from zodbpickle.pickle import Pickler
-
-from ZEO._compat import Unpickler
+from zodbpickle.pickle import Unpickler
 
 
 class TransactionBuffer:
--- a/src/ZEO/TransactionBuffer.py
+++ b/src/ZEO/TransactionBuffer.py
@@ -23,7 +23,8 @@ is used to store the data until a commit or abort.
 
 import tempfile
 
-from ZEO._compat import Pickler
+from zodbpickle.pickle import Pickler
+
 from ZEO._compat import Unpickler
 
 
--- a/.meta.toml
+++ b/.meta.toml
@@ -32,10 +32,8 @@ testenv-setenv = [
 [flake8]
 additional-config = [
     "# E501 line too long",
-    "# F401 unused import",
     "per-file-ignores =",
     "    src/ZEO/tests/test_cache.py: E501",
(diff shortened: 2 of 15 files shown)


src/ZEO/TransactionBuffer.py
```diff
@@ -23,7 +23,8 @@
 
 import tempfile
 
-from ZEO._compat import Pickler
+from zodbpickle.pickle import Pickler
+
 from ZEO._compat import Unpickler
```
d-maurer, line 28: We could use the standard `Unpickler` here.

dataflake (approved):

dataflake:
If you want to make a 6.0 release the pypy test failures should be fixed first.

d-maurer (approved):

d-maurer:
The failing PyPy3 checks result from a buggy `IterationTest.checkIteratorGCStorageTPCAborting` test. The test creates an iterator causing an entry in the `WeakValueDict` `_iterators` and the `set` `_iterator_ids`. The iterator reference is not stored; this means that the iterator can be garbage collected immediately. The test failure is caused because `_iterators` has less elements than `iterator_ids` which happens when there has meanwhile been a garbage collection. `_assertIteratorIdsEmpty` tries to prevent such a race condition by a call to `gc.disable()` but this can be too late. A fix would be to disable garbage collection before the iterator is created.

src/ZEO/TransactionBuffer.py
perrinjerome: This is because `noload` is not used here, right ?

d-maurer: The only difference between `zodbpickle.pickle.Unpickler` and `_compat.Unpickler` is the possibility to override `find_global`. This is used in `asyncio.marshal` but not in `TransactionBuffer`.

perrinjerome: ah I see, I thought by "standard `Unpickler`" you meant `pickle.Unpickler`, but you meant `zodbpickle.pickle.Unpickler`.

tseaver: Why haven't we updated `zodbpickle.Unpickler` to include the `find_global` support, rather than subclassing it here?

icemac:

